PR #6, opened 2026-04-17
WIP: document discrete-adjoint QH benchmark window

rogeriojorge:
## Summary
- logs the current stable QH comparison window against classic VMEC2000 in the discrete-adjoint branch plan
- documents that the branch is now derivative-stable enough for production-path benchmarking, but still blocked on exact forward solve cost and memory retention

## Current state
The branch has already fixed the late-iteration Jacobian failure that previously produced `O(1e19)` optimality values on QH. The remaining blocker is no longer derivative correctness on the tested window; it is exact forward runtime and RSS on nearby boundary points.

## Why draft
This PR is draft because the branch still needs one more forward-path/runtime pass before it is ready to merge.


rogeriojorge:
Closing for now. The discrete-adjoint branch still needs another forward-path/runtime pass before the QH VMEC-JAX benchmark is competitive with the classic reference.

PR #47, opened 2026-07-19
Drift-gate the R25.3 recycled-Jacobian carry

rogeriojorge:
Closes the loop on the plan-R25.3 recycled implicit-Jacobian lane using the new solvax 0.8.7 drift diagnostic. Additive and capability-guarded: behavior is unchanged on older solvax, and unchanged everywhere unless `recycle=True` is already in use.

## Context (why this lane was stuck)
R25.3 threads one GCROT deflation pair `(C, U)` through all `2*nm` per-dof implicit solves — `lax.scan` over dof chunks, vmapped within a chunk, lane-0 pair advanced — and stashes it between `jac` calls (the R18b shared-solver consolidation made all of these `solvax.gcrot`). It stayed **opt-in/EXPERIMENTAL** for a measured reason documented in `jacobian_rows_recycled` and the `least_squares` note: after a large accepted trust-region step the operator `Fz` has moved, the stashed pair no longer deflates it, and *the recycle space measurably slowed warm-started columns* — the Item I.8c A/B question. The lane had no way to know a pair had gone stale.

## What this adds
solvax ≥ 0.8.7 (`recycle_drift`, uwplasma/SOLVAX#32) measures exactly that staleness on every warm start, inside the existing warm-start QR at **zero extra matvecs**: the mean principal-angle sine between the incoming recycle image space and its re-established span, with the gap-free bound `sinΘ ≤ ‖ΔA‖‖U‖` (linear in the trust-region step; measured ~1e-3/step on slowly-varying operators, ratio 10.0 per decade of step size).

- `implicit._recycled_solve` now surfaces the diagnostic (probe `_GCROT_REPORTS_DRIFT`, same capability idiom as `_SCHUR_ACCEPTS_D_BLOCK`).
- `optimize.jacobian_rows_recycled` **drift-gates the carry**: lane-0 drift beyond `_RECYCLE_DRIFT_LIMIT = 0.5` (an essentially rotated-away space) drops the pair to a cold start for the next chunk — a stale pair can now slow at most the one chunk that measures it, directly addressing the I.8c slowdown mechanism. Max drift over chunks is returned, stashed in the holder, and printed under `verbose`.

## Intrusiveness
- No new dependencies (`solvax` already core); no API changes; `recycle=False` paths untouched.
- Older solvax: probe is False → identical behavior to today (unconditional carry).
- Within one `jac` call the operator is fixed (measured drift ~1e-16), so the gate is inert there; it acts precisely at the stash-across-iterates boundary R25.3 was designed for.

## Tests
`tests/test_recycle_drift.py` pins the vmex-side contract on a synthetic pytree operator through `_recycled_solve`: cold start drift 0, unchanged operator < 1e-10, monotone growth with the operator step, and well under the gate at slow-drift steps. Skips cleanly on pre-0.8.7 solvax. (The full recycled lane remains gated behind the I.8c A/B as planned; this PR gives that measurement its missing instrument.)

### DIFF
--- a/tests/test_optimize_penalty.py
+++ b/tests/test_optimize_penalty.py
@@ -152,3 +152,22 @@ def test_implicit_lane_jac_fallback_and_diagnostic_resolve(monkeypatch, capsys):
     assert np.isfinite(res.cost)
     assert res.equilibrium is not None  # cold-solve fallback delivered it
     assert res.equilibrium.result.converged
+
+
+def test_recycled_jacobian_lane_scan_shapes() -> None:
+    """The drift-gated recycled Jacobian lane (recycle=True) must run.
+
+    Exercises optimize.jacobian_rows_recycled's lax.scan carry — the path that
+    threads and drift-gates the GCROT pair across dof chunks. A shape bug there
+    (gating against the full stacked recs leaf instead of a single lane) makes
+    lax.scan reject the carry-type mismatch; this pins that the lane completes
+    on a real (small) equilibrium and returns a usable result. The recycled
+    lane is otherwise opt-in and was previously uncovered in CI.
+    """
+    inp = VmecInput.from_file(DATA_DIR / "input.solovev")
+    res = opt.least_squares(
+        OBJECTIVE, inp, max_mode=1, jac="implicit", recycle=True,
+        max_nfev=3, verbose=0,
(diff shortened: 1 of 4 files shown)


rogeriojorge:

PR #68, opened 2026-07-21
bench: add CPU/GPU solve and gradient-parity audit

rogeriojorge:
## Summary

Provide a reproducible hardware-parity audit using VMEX's public device APIs instead of environment-variable platform selection.

## Main changes

- Update production profiling and GPU-matrix scripts to use `device=`.
- Add a JSON-producing CPU/GPU parity audit.
- Compare forward state and boundary gradients for MHD energy, magnetic well, Mercier stability, quasisymmetry, and quasi-isodynamicity.
- Record selected devices, cold/warm timing, state norms, scalar outputs, and relative differences.
- Add CPU-only behavior and argument-validation tests.
- Run the quick five-physics CPU/GPU audit in trusted manual GPU CI and upload
  its JSON timing/parity artifact; timing is recorded but never used as a gate.
- Exercise a real tracked-mgrid NESTOR solve for 60 iterations on CPU/GPU and
  assert placement of its potential, mode matrix, nonsingular source, and
  boundary-pressure cache.
- Report bounded benchmark convergence honestly and use measured warm-run
  convergence/iteration data instead of hard-coded success.
- Split the long full-physics core nightly into four disjoint alphabetical
  jobs and isolate the QA/QH/QP campaigns on fresh workers.

## Results

On two RTX A4000 GPUs without platform-selection environment variables:

| Quantity | CPU/GPU relative difference |
| --- | ---: |
| MHD-energy gradient | `5.68e-16` |
| Magnetic-well gradient | `1.21e-10` |
| Mercier gradient | `1.56e-12` |
| Quasisymmetry gradient | `1.94e-15` |
| Quasi-isodynamic gradient | `5.75e-14` |

- Forward-state maximum and relative L2 differences are both exactly zero in
  the current quick audit.
- CPU-only hosts run the CPU lane and explicitly report the unavailable cross-device comparison.
- Bounded NESTOR activates the vacuum solve on both devices; CPU/GPU state and
  residual relative differences are `7.3e-15` and `2.4e-12`, respectively.
- Fresh VMEC2000 fixed-boundary comparisons of `wb/rmnc/zmns/iota` are within
  `9.14e-12` across Solovev, CTH, and QH cases. The converged CTH free-boundary
  comparison is looser, with a worst reported relative difference of `1.244e-4`.
- VMEX warm time was lower on all 14 benchmark rows in this run (speedup
  `1.37-3.91x`, median `1.88x`). The LASYM row is a bounded nonconverged stress
[...]

PR #69, opened 2026-07-21
Fix indexed INDATA parsing and localize non-finite forces

rogeriojorge:
## Summary

This PR is now stacked directly on #68 so the full sequence can be reviewed and tested before merge; its own diff remains limited to the parser/non-finite/axis-recovery changes below.

This PR fixes and diagnoses the fixed- and free-boundary iteration-1 NaN reported by a collaborator, without using or exposing the collaborator's input:

- parse indexed one-dimensional VMEC namelist arrays with Fortran lower bounds, multivalue starts, array sections, and overlay semantics;
- stop on the first non-finite force evaluation instead of printing NaN through `NITER`;
- identify the first failing field, normalization, force, Fourier, residual, or preconditioner stage with shareable value-free codes;
- apply VMEC2000's initial-axis recovery before free-boundary vacuum/axis-current construction.

## What the collaborator's report proves

The private case fails identically with `LFREEB=T` and `LFREEB=F`, with and without `PRECON_TYPE='NONE'`, and with supplied, omitted, or VMEC2000-derived axis coefficients. The old diagnostic reported finite setup arrays and a valid initial Jacobian, but non-finite raw residuals on iteration 1.

That rules out NESTOR/vacuum activation, the optional 2-D preconditioner, and the supplied axis as the common cause. They all occur after (or outside) the fixed-boundary raw-force failure. The probable common cause is the normalization input to the raw force residuals.

## Reproduced VMEC2000/VMEX split and root cause

VMEX's namelist parser handled indexed 2-D boundary coefficients, but silently ignored indexed 1-D assignments such as:

```fortran
APHI = 0.0
APHI(2) = 1.0
```

VMEC2000 overlays `APHI(2)` on the initialized array. Old VMEX retained only the dense leading zero, so the effective toroidal-flux derivative, magnetic energy scale, and lambda scale became zero. The unnormalized forces remained finite, but multiplication by singular force norms made `FSQR`, `FSQZ`, and `FSQL` NaN on the first pass. This mechanism is fully consistent with the collaborator's value-free `D03_RAW_FORCE_NONFINITE` report, although the private deck must be rerun to confirm that it is the exact case there.

A follow-up audit found a second legal Fortran form with the same failure:

```fortran
APHI(1) = 0.0, 1.0
```

VMEC2000 assigns both `APHI(1:2)` elements. VMEX had retained only the first zero. On the same public deck that interpretation reproduces finite setup and Jacobian, `WMHD=0`, and NaN in all three iteration-1 residuals. The parser now consumes consecutive values from an indexed starting element and handles one-dimensional Fortran array sections with inclusive bounds.

I reproduced the split with a public circular-tokamak deck. Local VMEC2000 and patched VMEX print the same first iteration to displayed precision:

```text
FSQR=2.00E+00  FSQZ=1.73E+00  FSQL=3.56E-02
RAX=6.000E+00  DELT=9.00E-01  WMHD=9.3751E+03
```

Simulating the pre-fix VMEX interpretation (`APHI=[0,0,...]`) on the same deck gives a zero energy scale and non-finite residuals. Both indexed forms now reproduce VMEC2000's displayed first iteration exactly. A second indexed-array VMEC2000 comparison reached machine-precision parity in `rmnc`, `zmns`, magnetic energy, and the `add_fluxes.f90` half-to-full iota reconstruction.

The parser now overlays indexed assignments for:

- `APHI`, `AM`, `AI`, and `AC` (including their different Fortran lower bounds);
- auxiliary profile arrays;
- `NS_ARRAY`, `FTOL_ARRAY`, and `NITER_ARRAY`, including VMEC2000's initialized gap values;
- `EXTCUR`;
- modern and obsolete indexed axis arrays.

## Privacy-safe diagnosis

`python tools/diagnose_input.py INPUT` evaluates one force pass and emits only runtime metadata, PASS/FAIL stages, and an assessment code. It never emits the input path/name, resolution, coefficients, profiles, flux/current values, axis values, force magnitudes, logs, or wout content unless the explicitly local-only `--details` flag is used.

The former broad `D03` is now localized to:

- magnetic-field assembly;
- volume, R/Z energy normalization, lambda scale, or lambda normalization;
- real-space force kernels, Fourier projection, or residual scaling;
- unnormalized force sums and individual normalized R/Z/lambda channels;
- preconditioner cache, RHS, radial solve, or lambda preconditioner.

Active reconstruction and RFP inputs are reported as unsupported before VMEX silently applies ordinary-equilibrium semantics. The reconstruction check follows `vsetup.f`/`read_indata.f`: an inert `LRECON=T` with no active MSE/Thomson data is not falsely rejected.

The stage checks are collected only by the diagnostic API; production solver iterations retain the existing finite-result guard without paying for all intermediate diagnostic reductions.

## Other behavioral fixes

- Every iteration-driving output is checked for NaN/Inf. A non-finite pass raises `VmecNumericalError` at iteration 1 with public `ier_flag=9`.
- VMEC2000 `eqsolve.f` retries a sign-changing initial Jacobian using `guess_axis`. VMEX now shares that recovery between fixed and free boundary and rebuilds the free-boundary axis-current/vacuum topology only after the retry.
- If the regenerated axis is still invalid, the normal typed Jacobian error is raised.

## Validation

Local checks after the new commit:

- parser/diagnostic/force focused suite: **108 passed**;
- fixed boundary, multigrid, free boundary, CLI, and solver suite: **157 passed, 12 skipped, 2 expected xfailed**;
- full implicit AD/FD suite with JIT: **19 passed**;
- Ruff over `vmex tests examples benchmarks tools`: clean;
- mypy over `vmex`: clean;
- `git diff --check`: clean.

After rebasing onto #68, the focused parser/non-finite/free-boundary group passes **98 tests with 1 expected missing-asset skip**; Ruff, mypy, and `git diff --check` remain clean. All **17** ordinary hosted checks pass on the rebased head, and GitHub reports the PR mergeable/clean.

After the multivalue/section follow-up, **111 parser, diagnostic, and validation tests pass** locally. The broader collaborator-path audit passes **186 tests with 9 optional skips**. Ruff and mypy remain clean.

All hosted CI shards pass, including the combined >=95% coverage gate.

## Confidentiality

No private input, filename, path, log, wout, coefficient, equilibrium value, mgrid, or coil data was received, copied, printed, or committed. All reproduction and parity work uses synthetic/public inputs. The requested collaborator rerun remains value-free and explicitly forbids `--details`.

### DIFF
--- a/tests/test_freeboundary.py
+++ b/tests/test_freeboundary.py
@@ -28,6 +28,7 @@ structural + coarse:
 
 from __future__ import annotations
 
+import dataclasses
 import re
 from pathlib import Path
 
@@ -395,6 +396,23 @@ def test_missing_mgrid_raises(tmp_path):
     inp = VmecInput.from_file(DECK)
     with pytest.raises(MgridNotFoundError):
         FB.solve_free_boundary(inp, mgrid_path=tmp_path / "mgrid_missing.nc")
+
+
+def test_bad_supplied_axis_is_reguessed_before_fused_filament(capsys):
+    """Free boundary must share fixed boundary's first-bad-axis recovery."""
+    inp = VmecInput.from_file(DECK)
+    raxis_c = inp.raxis_c.copy()
+    raxis_c[0] = 2.0  # deliberately outside the CTH-like plasma boundary
+    inp = dataclasses.replace(inp, raxis_c=raxis_c, niter_array=[2], nstep=1)
+    result = FB.solve_free_boundary(
+        inp, mgrid_path=MGRID, max_iterations=2, verbose=True,
+        error_on_no_convergence=False,
+    )
+    output = capsys.readouterr().out
+    assert "TRYING TO IMPROVE INITIAL MAGNETIC AXIS GUESS" in output
+    assert not result.converged
+    assert np.isfinite(result.fsqr)
+    assert result.r00 < 1.0
 
 
 def test_cli_missing_mgrid_fallback_warns(tmp_path):
--- a/tests/test_input_profiles.py
+++ b/tests/test_input_profiles.py
@@ -143,6 +143,38 @@ def test_indexed_aphi_overlays_dense_assignment() -> None:
     np.testing.assert_array_equal(inp.aphi[:3], [0.0, 1.0, 0.0])
 
 
+def test_indexed_aphi_starting_element_consumes_following_values() -> None:
+    """Fortran ``APHI(1)=0,1`` assigns two elements, not one scalar."""
+    inp = VmecInput.from_indata_text(
+        """&INDATA
+        MPOL = 3
+        NTOR = 0
+        APHI(1) = 0.0, 1.0
+        RBC(0,0) = 1.0
+        RBC(0,1) = 0.1
+        ZBS(0,1) = 0.1
+        /
+        """
+    )
+    np.testing.assert_array_equal(inp.aphi[:3], [0.0, 1.0, 0.0])
+
+
+def test_indexed_aphi_section_uses_fortran_inclusive_bounds() -> None:
+    """A non-leading array section retains VMEC's initialized first term."""
+    inp = VmecInput.from_indata_text(
+        """&INDATA
+        MPOL = 3
+        NTOR = 0
+        APHI(2:3) = 0.25, 0.5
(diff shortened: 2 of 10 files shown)


rogeriojorge:

PR #70, opened 2026-07-22
Complete VMEC2000 parity for free-boundary multigrid and solver recovery

rogeriojorge:
## Summary

Bring VMEX fixed- and free-boundary execution into explicit, tested parity with
the VMEC2000 driver and namelist semantics, while retaining VMEX extensions for
JAX execution, automatic differentiation, external-field adapters, and bounded
recovery.

This PR now includes:

- free-boundary radial multigrid with hot restarts and resolution-aware
  NESTOR/vacuum continuation;
- Fortran-compatible `&INDATA` array assignment replay;
- the VMEC2000 `LFORBAL=T` formulation on the correct radial meshes;
- VMEC2000-compatible `NS_ARRAY`, `FTOL_ARRAY`, `NITER_ARRAY`, axis-recovery,
  and residual-state transitions;
- robust but non-regularizing radial tridiagonal diagnostics through
  SOLVAX 0.8.8;
- best-finite-state JAC75 recovery for both fixed and free boundary;
- privacy-safe, stage-local numerical diagnostics;
- public stress tests and direct VMEC2000 trajectory/WOUT comparisons.

The branch contains no private input, input-derived value, filename, path,
solver log, coil/mgrid data, or WOUT. All committed validation cases use public
repository fixtures or generated data.

## Why this belongs in one PR

The affected behaviors meet at the first coarse solve and at every radial-grid
transition. Testing them independently can hide errors that only appear when a
Fortran namelist overlay, `LFORBAL`, axis recovery, a nonconverged coarse stage,
the ordinary radial preconditioner, and free-boundary vacuum state are combined.
The public 238-mode stress test deliberately exercises those seams together.

This is stacked on #69. Downstream PRs #71-#74 were created before some of the
follow-up corrections in this branch. Merge/retarget in stack order and refresh
each child from the updated parent. Conflict resolution must retain the
behaviors and tests listed below rather than taking an older downstream parser,
solver, CLI, free-boundary, or documentation file wholesale.

## VMEC2000 source contracts audited

The implementation was checked directly against the local VMEC2000/STELLOPT
sources, including:

- `readin.f`, `vmec_input.f`, and `runvmec.f`;
- `allocate_ns.f`, `initialize_radial.f`, and `interp.f`;
- `eqsolve.f`, `evolve.f`, `restart.f`, `funct3d.f`, and `residue.f90`;
- `fbal.f`, `add_fluxes.f90`, `precondn.f`, `scalfor.f`, and
  `tomnsp_mod.f`;
- the NESTOR/vacuum allocation, matrix, potential, and cadence paths.

The resulting contracts are documented in
`docs/input_reference.rst`, `docs/algorithms.rst`, and
`docs/vmec2000_compatibility.rst`.

### 1. Fortran namelist equivalence

`&INDATA` assignments are now replayed sequentially, using the declared
VMEC2000 bounds and Fortran element order (first subscript varies fastest).
This supports:

- dense writes followed by indexed/section overlays, and the reverse;
- whole and open sections, inclusive bounds, and positive/negative strides;
- starting-element sequence association such as `APHI(1)=...`;
- multidimensional starting-element writes;
- compact boundary sections such as `RBC(-6:6,0)=...`;
- repeated `NS_ARRAY`, `FTOL_ARRAY`, and profile assignments with source-order
  semantics.

Malformed or unsupported input still fails before the solve with a redacted
diagnostic classification.

### 2. Radial ladder parity

- `readin.f`'s positive nondecreasing `NS_ARRAY` prefix is used; the first
  decreasing or nonpositive entry terminates the ladder, while equal entries
  rerun.
- An all-zero `FTOL_ARRAY` is expanded with VMEC2000's logarithmic ladder.
- Partially assigned `NITER_ARRAY` entries retain `-1`; the solver performs the
  one first pass that `eqsolve.f` performs before the iteration-cap check.
- Increasing grids interpolate the preceding final `xc`, matching the
  `allocate_ns.f`/`initialize_radial.f` ordering.
- `fsqr/fsqz/fsql` continue across grids even though iteration counters reset.
  This is required for `residue.f90`'s first-fine-grid edge gate and was the
  remaining source of a large transition mismatch.

The same rules are used by fixed- and free-boundary multigrid.

### 3. Free-boundary multigrid and hot restart

- The external field is loaded/tabulated once per ladder.
- Increasing radial resolution rebuilds resolution-dependent NESTOR surface,
  Green-function, matrix/vector, potential, axis-current filament, and traced
  loop structures.
- Equal-grid stages may reuse valid dynamic caches.
- Vacuum activation occurs once; `ivac`, adaptive cadence, `delbsq`, evolved
  edge state, and constraint baselines continue across stages.
- An already-active fine stage starts from the carried edge state and performs
  its first new full vacuum update at the VMEC2000 cadence.
- User and internal hot restarts preserve the full plasma state and apply the
  documented reset/continuation semantics.
- Mgrid input and host-side ESSOS/SIMSOPT Biot-Savart protocols remain
  supported; the sampled field is represented by the common JAX mgrid path.

### 4. `LFORBAL` radial meshes

`LFORBAL=T` now uses the VMEC2000 averaged radial force-balance formulation in
the production force map, cache refresh, 2-D preconditioner map, diagnostics,
and WOUT processing. In particular, the effective half-mesh `chips` profile is
converted to the full-mesh `chipf` used by `calc_fbal`/`add_fluxes.f90`; using
half-mesh values directly produced a systematic physics and energy mismatch.

### 5. `PRECON_TYPE='NONE'`

VMEC2000's `NONE` disables the optional 2-D Newton/Krylov preconditioner. It
does **not** disable the ordinary 1-D radial and lambda preconditioners. VMEX
now preserves that distinction consistently in fixed boundary, free boundary,
multigrid, diagnostics, and documentation.

The radial solve remains unregularized. SOLVAX 0.8.8 adds per-column pivot and
backward-residual checks plus a deterministic identity fallback for a rejected
preconditioner column. VMEX reports
`D04E_RADIAL_PRECONDITIONER_REJECTED` instead of allowing a singular
preconditioner to inject NaN/Inf into an otherwise finite force evaluation.
SOLVAX PR
[uwplasma/SOLVAX#40](https://github.com/uwplasma/SOLVAX/pull/40) is merged and
the 0.8.8 release is published; VMEX now requires `solvax>=0.8.8`.

### 6. Axis and JAC75 recovery

- Production preserves a missing/all-zero supplied axis until VMEC2000's one
  `eqsolve.f` first-force recovery transfer; geometry-only test utilities may
  explicitly request pre-inference.
- The first bad-Jacobian/high-force recovery keeps the VMEC2000 momentum
  transfer and is available on every radial stage.
- When the original attempt reaches JAC75, bounded VMEX recovery restarts the
  best finite `xstore`, zeros momentum, reduces the initial `DELT`, retains the
  residual continuation, and does not replace the checkpoint with another
  `LMOVE_AXIS` cold reconstruction.
- Free-boundary JAC75 recovery rebuilds all geometry-dependent NESTOR/vacuum
  structures rather than reusing failed-attempt caches.
- `jacobian_retries=0` retains the exact immediate VMEC2000 fatal-stop policy.

`LFULL3D1OUT=T` continues to permit WOUT output after ordinary `NITER`
exhaustion, but not after a fatal numerical/Jacobian abort.

### Combined 238-mode fixed-boundary stress

`tests/test_vmec2000_feature_stress.py` constructs a public 238-mode deck from
a tracked boundary and combines:

- no supplied axis;
- `LFORBAL=T`;
- `PRECON_TYPE='NONE'`;
- APHI starting-element sequence association;
- repeated dense/indexed `NS_ARRAY` and `FTOL_ARRAY` writes;
- compact multidimensional RBC/ZBS sections;
- the complete `21 -> 34 -> 55 -> 89` radial ladder.

Against the local VMEC2000 Release executable, VMEX reproduces the displayed
trajectory exactly:

- stage exits at iterations 462, 139, and 543;
- identical first rows after each interpolation;
- two Jacobian resets on the finest stage;
- identical residual screen row at iteration 10,000, where the intentionally
  extreme final tolerance is not reached by either code.

The quick test pins the converged first stage; a full/nightly test pins all four
stages and the ordinary final iteration-limit state.

### Public converged free-boundary benchmark

The tracked CTH-like `7 -> 15` mgrid benchmark now has the same one vacuum
activation and the same first fine-grid residual row as VMEC2000. Both codes
converge the fine stage at iteration 340. Final normalized WOUT differences are:

- `rmnc`: `6.08e-5`;
- `zmns`: `3.59e-4`;
- `iotaf`: `1.99e-6`;
- magnetic energy: `6.07e-8` relative.

The benchmark report embeds only aggregate timings/residuals and normalized
errors, never input or field data.

### Current local gates

- ordinary suite: **757 passed, 34 skipped, 2 expected xfails**;
- combined 238-mode public stress: **3 passed**, including the exact full
  four-stage trajectory;
- full free-boundary physics selection: **3 passed**, covering active 2-D
  preconditioning, `LFORBAL` across vacuum activation, and converged WOUT
  parity;
- Ruff: clean;
- mypy: clean across 53 source files;
- Sphinx documentation with warnings as errors: passed;
- wheel and source distribution builds: passed;
- `git diff --check`: clean.

### Final hosted validation

The original catch-all parity-C worker was externally shut down twice above
84% with no failed assertion or traceback. Its 41-file ownership set is now
partitioned explicitly into disjoint C1/C2 shards; a mechanical ownership
check proves that all 53 ordinary core test files remain assigned exactly once,
and the coverage gate combines both new artifacts.

[CI run 30111998499](https://github.com/uwplasma/vmex/actions/runs/30111998499)
passed all **18** ordinary checks, including C1 (16m06s), C2 (6m19s), all
three implicit-gradient shards, and the combined **>=95%** coverage gate. The
only skipped job is the workflow's expected nightly/manual full-physics matrix.

## Merge-preservation checklist

The final merged stack must retain:

- sequential Fortran namelist replay and compact multidimensional boundary
  sections;
- privacy-safe first-force and radial-preconditioner diagnostics;
- fixed/free complete ladder and residual-state continuation;
- final-`xc` interpolation and equal/decreasing stage semantics;
- correct full-mesh `LFORBAL` data flow;
- `PRECON_TYPE='NONE'` as ordinary 1-D preconditioning, not no
  preconditioning;
- one-transfer axis recovery and best-state JAC75 recovery;
- WOUT-on-ordinary-limit behavior under `LFULL3D1OUT`;
- free-boundary hot restart, one-time vacuum activation, cadence/edge/constraint
  continuation, and resolution-aware NESTOR rebuilds;
- mgrid and ESSOS/SIMSOPT external-field adapters;
- free-boundary AD-versus-FD coverage;
- the public combined stress case and VMEC2000 benchmark;
- the compatibility/design documentation added here.

Post-merge validation should run the parser/diagnostic suite, fixed/free
multigrid tests, the mandatory public stress diagnostic, and the converged
public free-boundary benchmark. The full four-stage stress trajectory belongs
in nightly/manual full-physics CI.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -20,10 +20,10 @@ env:
 permissions:
   contents: read
 
-# Sharded CI (plan.md Phase-2 review item 1): five parallel test jobs, each
-# well under the 10-minute wall-time budget, plus a fast lint gate (ruff +
-# mypy) and a coverage gate that combines the parity + gradient shards and
-# enforces >= 95% on vmex.
+# Sharded CI (plan.md Phase-2 review item 1): bounded parallel test jobs, plus
+# a fast lint gate (ruff + mypy) and a coverage gate that combines the parity
+# and gradient shards and enforces >= 95% on vmex.  The explicit job limits
+# include dependency setup, coverage handling, and hosted-runner variance.
 jobs:
   lint:
     # Fast fail-first gate (plan Item I.3): ruff + mypy over the whole tree,
@@ -181,8 +181,8 @@ jobs:
           print("golden dir:", mod.resolve_golden_dir())
           EOF
 
-      # Four shards, each well under the 10-min budget.  The xdist dist MODE is
-      # chosen PER SHARD by whether its modules share an expensive fixture:
+      # Four bounded shards.  The xdist dist MODE is chosen PER SHARD by
+      # whether its modules share an expensive fixture:
       #   * loadscope groups a whole module onto ONE worker, so module-scoped
       #     equilibrium fixtures (omnigenity qi_eq/tok_eq, bootstrap eq,
       #     stability vacuum/highbeta/shaped) are solved ONCE, not once-per-
@@ -251,12 +251,14 @@ jobs:
   gradient:
     name: Implicit-gradient suite ${{ matrix.shard }} (JIT on, coverage)
     runs-on: ubuntu-latest
-    # Sharded so no single job exceeds the timeout even with a COLD finite-
-    # difference cache (first run after the test file changes recomputes every
-    # FD reference). The two dominant FD tests (lasym-adjoint, lasym-3d) each get
-    # their own shard; shard "c" is the 17 remaining, lighter tests. Each shard
-    # keeps its own FD reference cache. Warm-cache runs finish in ~4 min.
-    timeout-minutes: 15
+    # Sharded so a COLD finite-difference cache remains bounded (first run
+    # after the test file changes recomputes every FD reference).  The two
+    # dominant FD tests (lasym-adjoint, lasym-3d) each get their own shard;
+    # shard "c" owns the remaining tests.  A slow hosted runner measured
+    # 14m19s for those tests alone, so include installation, coverage upload,
+    # and runner variance rather than canceling a fully passing test process
+    # during post-processing.  Warm-cache runs remain much faster.
+    timeout-minutes: 25
     strategy:
       fail-fast: false
       matrix:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -148,7 +148,7 @@ jobs:
     strategy:
       fail-fast: false
       matrix:
-        shard: [a, b, c, d]
+        shard: [a, b, c1, c2, d]
     steps:
       - uses: actions/checkout@v6.0.2
 
@@ -181,7 +181,7 @@ jobs:
           print("golden dir:", mod.resolve_golden_dir())
           EOF
 
-      # Four bounded shards.  The xdist dist MODE is chosen PER SHARD by
+      # Five bounded shards.  The xdist dist MODE is chosen PER SHARD by
       # whether its modules share an expensive fixture:
       #   * loadscope groups a whole module onto ONE worker, so module-scoped
       #     equilibrium fixtures (omnigenity qi_eq/tok_eq, bootstrap eq,
@@ -201,7 +201,9 @@ jobs:
       #   a = heaviest solver-golden modules, including all free-boundary
       #       modules so their JAX caches leave with this bounded shard instead
       #       of accumulating in the catch-all worker processes;
-      #   c = everything else, incl. omnigenity/turbulence/bootstrap/stability.
+      #   c1/c2 = the former catch-all, split explicitly after two hosted
+      #           runners were externally shut down above 84% despite no test
+      #           failure.  Explicit file ownership keeps coverage disjoint.
       # Together they cover the whole non-gradient, non-example core suite once;
       # the separate mirror matrix owns tests/mirror. Therefore
       # every test file appears in exactly one shard (coverage gate combines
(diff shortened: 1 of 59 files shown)


rogeriojorge:
## Follow-up: Fortran multidimensional INDATA sections

A collaborator privacy-safe rerun exposed a parser compatibility gap at input-read time: legal VMEC namelist sections such as `RBC(-6:6,0) = ...` were rejected. No collaborator input, path, coefficient, or equilibrium value was used or copied here.

### Root cause

This is unrelated to NetCDF. VMEC2000 reads the text `&INDATA` block with a native Fortran namelist. VMEX uses a handwritten host-side parser. Historically, that parser treated every key containing `:` as an unindexed dense assignment; the boundary loader then ignored that assignment, allowing boundary sections to be silently lost. The parser hardening correctly replaced the silent loss with a `ValueError`, but did not yet implement multidimensional sections.

### Fix in `a73ca47d`

- expand explicitly bounded sections in one or more dimensions, with inclusive Fortran bounds and positive or negative strides;
- populate values in Fortran column-major array-element order, verified against a compiled Fortran namelist probe;
- support compact `RBC`, `ZBS`, `RBS`, and `ZBC` sections through the common indexed map;
- retain the existing one-dimensional `APHI`, profile, and multigrid semantics;
- report `D00C_INPUT_PARSE_ERROR` without echoing private syntax when another parse error remains.

### Validation

- parser/nonfinite diagnostic slice: **99 passed**;
- integrated fixed/free-boundary, multigrid, AD/FD, CLI, and preconditioner slice: **132 passed, 3 optional skips**;
- a section-only public boundary reaches `OK_FIRST_FORCE_PASS_FINITE`;
- a public QI diagnostic deck containing the structural form `RBC(-6:6,0)` reaches `OK_FIRST_FORCE_PASS_FINITE`;
- Ruff and `git diff --check`: clean.

This may explain the original first-iteration NaNs more directly than `APHI`: before fail-fast parsing, a compact boundary section could be silently discarded, potentially including the major-radius coefficient.

rogeriojorge:

PR #76, opened 2026-07-23
Accelerate high-mode solves and bound implicit Jacobian storage

rogeriojorge:
## Summary

This PR addresses the collaborator-reported high-resolution VMEX runtime and
memory problem with measured, hardware-specific changes. It replaces the
dominant full mode-stacked forward synthesis with a separable toroidal FFT on
hardware where that kernel wins, releases completed radial-stage executables
in one-shot CLI runs, and prevents scalar implicit objectives from assembling
the much larger forward block Jacobian.

The physics, convergence criteria, public device contract, dense expert
helpers, and free-boundary/NESTOR path are unchanged.

## Goals

- reduce time and peak RSS for the supplied 858-mode HSX case;
- preserve VMEC2000 equilibrium/WOUT accuracy and CPU/GPU derivatives;
- avoid environment-variable hardware routing;
- bound implicit differentiation according to derivative direction;
- retain explicit fallbacks and avoid regressions for routine x86 users;
- identify the exact remaining work needed for VMEC++-class memory and
  high-mode vector diagnostic gradients.

## Root causes

Two independent costs were being conflated:

1. The forward solver expanded every signed Fourier mode over the full
   `(theta, zeta)` grid. The supplied deck required about 17.8 billion
   compiler-estimated FLOPs per synthesis.
2. Vector implicit differentiation assembles and factors exact radial block
   bands with `O(ns * m_block^2)` storage. At 858 modes, the active
   three-field block is 2574 square; a probe tensor or one band is already
   about 16 GiB. Column chunking cannot remove that quadratic factor storage.

VMEC++ avoids the first cost with a per-surface batched toroidal transform and
short poloidal accumulation. SOLVAX's generated block factor avoids one input
copy but still retains all Schur factors/off-diagonal bands, so a genuinely
lower-storage vector solve needs generated block rows plus checkpointed
elimination or a stronger matrix-free preconditioner.

### Measured forward synthesis

- repack signed helical coefficients into separable poloidal/toroidal blocks;
- use `jax.numpy.fft.irfft` for unaliased toroidal synthesis, with the existing
  dense DFT as the undersampled fallback;
- select FFT synthesis only above 512 modes on accelerators and ARM CPUs;
  smaller problems and x86 CPUs retain dense real contraction;
- preserve explicit `use_fft=True` / `False` overrides and active
  `jax.default_device(...)` contexts;
- keep low-level helpers dense by default and leave free-boundary/NESTOR
  transforms unchanged;
- clear only JAX's in-memory compilation/staging caches between distinct
  radial grids in the one-shot fixed-boundary CLI. The persistent disk cache
  is retained, and library `solve_multigrid` calls keep warm executables by
  default.

### Direction-adaptive implicit differentiation

- make `jac_solver="auto"` choose one exact reverse adjoint for a scalar
  residual and the established block path for vector residuals;
- retain explicit `reverse`, `block`, and `gmres` controls;
- cap device-aware automatic probe widths by `ceil(sqrt(dim))`, preventing a
  large accelerator budget from expanding to one full batch (the observed
  failure requested 1.16 TB);
- keep the dense-real transform for implicit residuals because direct complex
  FFT tangents and mixed fast/dense executables exceeded 7–10 GiB.

### Tests and documentation

- add value/JVP/VJP parity tests for symmetric and LASYM transforms, including
  Nyquist-extended WOUT grids;
- test measured kernel defaults, explicit overrides, and active JAX contexts;
- test scalar reverse/block/GMRES Jacobian parity and bounded auto chunks;
- test opt-in multigrid cache release and CLI selection;
- update README, changelog, algorithm, optimization, and performance docs.

## Results

All supplied-HSX runs converged in the same 2737 final-stage iterations.

| host / lane | before | this PR | result |
|---|---:|---:|---|
| Apple M4 CPU | 676.46 s / 3.896 GiB | **206.56 s / 1.634 GiB** | 3.27x faster, 58.1% lower RSS |
| Xeon W-2295 CPU | 426.94 s / 6.52 GiB | **413.24 s / 4.04 GiB** | 3.2% faster, 38.0% lower RSS |
| RTX A4000 | 1468.07 s / 5.40 GiB | **1166.47 s / 3.67 GiB** | 20.5% faster, 32.0% lower host RSS |

The rejected universal-FFT x86 candidate took 570.47 s, which is why this PR
uses a measured hardware default instead of changing every CPU to FFT.
On the M4, FFT was 38–88% slower warm on three routine 5–8-mode decks, only
8% faster warm (and 13% slower on the first in-process solve) at 128 modes.
Both lanes reached the supplied 162-mode Nuhrenberg–Zille NITER cap, so that
is not a convergence discriminator; dense was nevertheless 4.3% faster
(279.55 s versus 291.69 s). The automatic threshold therefore starts above
512 modes; explicit `use_fft=True` remains available for research.

Same-host controls on the exact input:

- VMEC++: 449.79 s / 445.1 MiB (one thread);
- VMEC++: 92.03 s / 380.0 MiB (ten threads);
- VMEC2000: 1154.82 s / 265.0 MiB.

VMEX is now faster than the measured one-thread VMEC++ control on the M4, but
ten-thread VMEC++ remains 2.24x faster and VMEX still uses about 4.4x its RSS.
The collaborator's forward runtime problem is substantially improved; the
parallel-runtime and storage gaps are not claimed closed.

Against the clean VMEC2000 WOUT, the final M4 relative-L2 errors are
`5.96e-12` (`rmnc`), `3.42e-11` (`zmns`), `9.72e-12` (`bmnc`), and
`2.37e-11` (core iota). The final x86 errors are `1.02e-11`, `6.57e-11`,
`1.75e-11`, and `4.10e-11`, respectively.

The exact high-mode scalar reverse Jacobian completed finite (`1 x 120`, norm
`65.5595641737`) in 993.76 s / 12.55 GiB. This avoids the vector block path
that grew beyond 39 GiB, but it is not yet VMEC++-class memory.

The exact factor-storage follow-up is now implemented separately in
[SOLVAX PR #39](https://github.com/uwplasma/SOLVAX/pull/39): at
`N=128, m=64` it lowers compiled temporaries by 90.4% on the M4 and 82.4% on
the A4000 with roundoff-identical primal/transpose/JVP/VJP results. VMEX cannot
adopt it mechanically because the current global colored-JVP tensor is already
about 16 GiB before factorization; a row/segment derivative generator remains
the integration requirement.

## Validation

Completed before publication:

- complete local suite: **707 passed, 88 skipped, 2 xfailed**;
- focused final transform/device/multigrid/CLI suite: **49 passed, 2 xfailed**;
- free boundary, LASYM NESTOR, multigrid, and hot restart:
  **32 passed, 6 skipped, 2 xfailed**;
- examples, package API, and CLI: **44 passed, 20 expected skips**;
- live opt-in VMEC2000 integration: **3 passed**;
- final office GPU focused suite: **55 passed, 1 skipped**, with no JAX
  platform-selection environment variables;
- strict Sphinx, Ruff, mypy, and `git diff --check`: passed;
- exact Apple CPU, office x86 CPU, and RTX A4000 supplied-HSX solves: passed;
- reverse/block/GMRES parity and frozen-path finite-difference diagnostic AD:
  passed.
- final source-isolated local full/nightly marker suite:
  **50 passed, 2 skipped** in 4006.28 s;
- post-push required GitHub CI:
  **17 passed** ([run 30072132561](https://github.com/uwplasma/vmex/actions/runs/30072132561)).

The seven-metric CPU/GPU gradient audit passes:

| gradient | relative difference |
|---|---:|
| MHD energy | `7.57e-16` |
| magnetic well | `1.50e-13` |
| DMerc | `1.03e-11` |
| `jdotb` | `1.97e-12` |
| Glasser `D_R` | `1.86e-11` |
| quasisymmetry | `4.55e-15` |
| quasi-isodynamicity | `1.73e-16` |

The post-push manual full/nightly workflow also passed all **32 jobs**
([run 30072167281](https://github.com/rogeriojorge/vmec_jax/actions/runs/30072167281)).
The trusted GPU workflow cannot be dispatched from this stack until its
workflow file reaches the repository default branch; the direct two-A4000
suite and seven-gradient audit above provide the current GPU gate without
environment-variable routing.

## Compatibility and user impact

- Existing calls remain source-compatible.
- Automatic device placement, explicit platform names/JAX devices,
  `device=None`, and active default-device contexts retain their documented
  meanings.
- No hardware-selection environment variable is required.
- x86 CPU users retain the faster dense kernel automatically.
- Repeated library solves retain warm executables unless the caller explicitly
  opts into stage-cache release.
- Free-boundary, NESTOR, mirror, LASYM, WOUT, and diagnostic behavior is
  unchanged by the forward transform selection.

## Remaining work

This PR deliberately does not claim to finish every performance goal:

1. fuse the inverse force transform (poloidal reduction plus batched `rfft`);
2. merge the exact checkpointed factorization in SOLVAX PR #39, then generate
   VMEX radial rows/segments without retaining the global colored-JVP tensor;
3. strengthen the matrix-free adjoint with a low-mode angular coarse space
   plus the VMEC radial solve;
4. evaluate masked single-executable multigrid only if it lowers peak RSS
   without penalizing scans;
5. keep compressed angular Schur complements as a guarded research lane with
   condition/pivot/error diagnostics and an exact fallback;
6. independently validate finite-current `jdotb`, DMerc components, and
   Glasser `D_R` base values before broadening LASYM diagnostic optimization.

The source audit, rejected experiments, numerical design, acceptance gates,
and follow-on ordering are maintained in
[`rogeriojorge/vmex-plans`](https://github.com/rogeriojorge/vmex-plans/blob/master/notes_r27_performance_methods.md).

## Final forward-integrated validation (2026-07-24)

The validated downstream tree passed the complete local suite (**753 passed, 92 skipped, 2 xfailed**) and the ordinary-discovery office A4000 gate (**10 passed**) with both GPUs visible and no JAX platform-selection environment variable. Two fresh full supplied-HSX runs completed in **224.64 s / 1.708 GiB** and **218.94 s / 1.569 GiB** and produced byte-identical WOUTs. Relative-L2 differences from VMEC2000 were `5.96e-12` (`rmnc`), `3.42e-11` (`zmns`), `1.02e-11` (`bmnc`), and `2.37e-11` (core iota).

A final set of exact candidates was measured and rejected rather than added: reverse rematerialization increased memory by 14--18%; global and local generated-row checkpointing were 3.7x and 1.96x slower with higher peaks; the existing two-dimensional Newton--Krylov lane did not beat VMEC stepping; and larger donated CLI scan blocks were slower and larger. The retained PR is therefore the smallest measured improvement. Closing the remaining ten-thread VMEC++ gap requires a fused allocation-free transform/force backend and threading, not another small JAX graph rearrangement.

The CI follow-ups change workflow partitioning only. Current performance head `d05bb69f` passed all **18 required jobs**, including the ≥95% coverage gate ([run 30101409736](https://github.com/uwplasma/vmex/actions/runs/30101409736)). Current integrated head `697d3e33` differs from validated source head `a7d52cca` only in `.github/workflows/ci.yml`.

Final exact-head manual matrix passed all **34 jobs** ([run 30101437203](https://github.com/rogeriojorge/vmec_jax/actions/runs/30101437203)), including the bounded core/free-boundary split and the 89-minute mirror-beta observables shard.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -35,6 +35,22 @@ follow [Semantic Versioning](https://semver.org/spec/v2.0.0.html) once past 1.0.
   implicit-gradient parity without JAX platform environment variables.
 
 ### Changed
+- **Measured high-mode synthesis.** Above 512 modes on accelerators and ARM
+  CPUs, equilibrium solves now repack signed helical modes into toroidal FFTs
+  plus a short poloidal contraction; smaller decks retain the dense lane
+  after routine M4 timing. The exact supplied HSX run on an Apple M4 falls
+  from
+  676.46 s / 3.896 GiB to
+  206.56 s / 1.634 GiB without changing convergence or WOUT parity. The
+  dense real contraction remains the x86 CPU default after it beat the FFT
+  path 413.24 s to 570.47 s on the office Xeon; explicit `use_fft` wins.
+  The one-shot CLI releases completed radial-stage executables; library
+  calls retain them by default. Implicit AD keeps the established real
+  contraction to avoid complex batched tangent storage.
+- **Direction-adaptive implicit Jacobians.** Scalar least-squares residuals
+  now use one matrix-free reverse adjoint by default, avoiding the dense
+  high-mode block assembly; vector residuals retain the exact block path.
+  Automatic probe chunks are capped by the conservative square-root width.
 - **Mirror-specific device policy.** Fixed/free-boundary mirror solves and
   beta scans now accept the common `device=` contract and default to CPU for
   their measured host-SciPy/JAX callback workload.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ follow [Semantic Versioning](https://semver.org/spec/v2.0.0.html) once past 1.0.
 ## [Unreleased]
 
 ### Added
+- **High-resolution resource profiler.**
+  `benchmarks/profile_high_resolution.py` records fresh-process implicit
+  Jacobian checksums and per-resolution mirror wall/RSS scaling without
+  hardware-selection environment variables.
 - **Explicit device selection.** Forward, free-boundary, implicit, and CLI
   solve paths accept public CPU/GPU/JAX placement controls: omitted or
   `"auto"` retains VMEX's measured policy, `None` follows JAX, and explicit
--- a/README.md
+++ b/README.md
@@ -164,15 +164,33 @@ CPU, single thread; `benchmarks/baseline.json`; reproduce with
   858-mode HSX case was instead 3.44× slower than CPU. The measured device
   policy therefore keeps both small-work and very-high-mode stages on CPU,
   while explicit placement always wins.
-- **Memory** — the bundled warm cases peak at 0.6–3.3 GB, but that is not a
-  high-resolution upper bound. The supplied `ns=101`, `mpol=18`, `ntor=24`
-  HSX deck used 3.90 GiB in a fresh CPU VMEX process versus 380 MiB in VMEC++
-  and 265 MiB in one-radial-process VMEC2000.
+- **High resolution** — on Apple Silicon, separable toroidal FFT synthesis
+  cuts the supplied
+  `ns=101`, `mpol=18`, `ntor=24` HSX deck from 676.46 s / 3.90 GiB to
+  206.56 s / 1.63 GiB in a fresh CPU VMEX CLI process; the CLI releases
+  completed radial-stage executables while the reusable library API retains
+  them by default. Convergence and VMEC2000 parity are unchanged. That is
+  faster than the measured
+  one-thread VMEC++ control (449.79 s), though ten-thread VMEC++ remains
+  faster and smaller (92.03 s / 380 MiB); one-radial-process VMEC2000
+  took 1154.82 s / 265 MiB.
+  The measured default selects this kernel only above 512 modes on ARM CPUs
+  and accelerators. Smaller problems retain the established dense lane (FFT
+  was slower on three routine M4 decks and only marginally faster warm at
+  128 modes); x86 CPUs also remain dense because FFT was slower.
+  Dense synthesis with CLI stage-cache release took 413.24 s /
+  4.04 GiB on the office Xeon, versus 570.47 s for FFT and the previous
+  cache-retaining dense baseline of 426.94 s / 6.52 GiB (3.2% faster and
+  38.0% lower peak RSS). `use_fft=True` or `False` remains an explicit
+  library override.
   Column chunking bounds simultaneous design probes, while the implicit
   block factors still scale as `O(ns * m_block²)`; an `ns=201` one-Jacobian
-  measurement used 4.11 GiB. A candidate chunk schedule that raised RSS by
-  36.7% was rejected, and the lower-memory GMRES path was over 5× slower
-  without finishing. Reducing the block-factor storage remains open work.
+  measurement used 4.11 GiB. The implicit callback deliberately retains the
+  real dense transform: complex FFT tangents exceeded 10 GiB. A candidate
+  chunk schedule that raised RSS by 36.7% was rejected, and the lower-memory
+  GMRES path was over 5× slower without finishing. Scalar objectives now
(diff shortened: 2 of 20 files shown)


rogeriojorge:

PR #80, opened 2026-07-25
Consolidate all open work into main: high-resolution performance, free-boundary parity, device/GPU stack

rogeriojorge:
PR #80 is the consolidation milestone for the VMEX work completed before the later #81/#82 branches. It is intended to make `main` installable, predictable on CPU/GPU, substantially faster, and evidence-backed against VMEC2000 without claiming research features that are not implemented.

## Goals

- Preserve VMEC2000-compatible fixed- and free-boundary physics, multigrid, restart, NESTOR/vacuum, LASYM, WOUT, and failure semantics.
- Let normal JAX hardware discovery choose CPU/GPU without requiring VMEX-specific platform environment variables.
- Make implicit differentiation and physics objectives reliable on CPU, the default GPU, and a non-default GPU.
- Bring the supplied high-mode HSX QHS runtime to about 2.5x ten-thread VMEC++ while keeping the remaining memory gap explicit.
- Consolidate the open work into one reviewable branch with packaging, docs, examples, benchmarks, and sharded release gates.

### Device and accelerator contract

- Low-level implicit APIs now use ordinary JAX placement by default: omitted `device` and `device=None` do not pin the differentiable parameter pytree to CPU.
- The high-level optimizer retains its measured `AUTO` policy, so this does not silently force every optimization onto a GPU.
- Explicit CPU/default-GPU selection is honored. A non-default accelerator uses the public `device_scope(device)` around parameter creation and differentiation; an unsupported explicit-without-scope call now raises clearly instead of returning plausible but wrong gradients.
- The returned concrete convenience runtime is rehomed to the parameter pytree's single committed accelerator without constraining the traced residual/adjoint graph.
- Mirror solves keep their separate measured host/SciPy policy. VMEX does not inject or require `JAX_PLATFORMS`, `JAX_PLATFORM_NAME`, `CUDA_VISIBLE_DEVICES`, or `XLA_FLAGS` for hardware discovery.

### Fixed/free-boundary solver and NESTOR

- Fixed and free multigrid, hot restarts, Jacobian recovery, LASYM NESTOR WOUT fields, mgrid compatibility, dense/FFT routing, and VMEC2000 output gating are covered by public regression campaigns.
- Production NESTOR now factors the dense mode matrix once with LU and reuses the factors for skip updates, following the safe LAPACK-style structure used by VMEC2000/VMEC++ rather than repeating `jnp.linalg.solve`.
- On GPU free-boundary runs, only NESTOR's small dense factor/solve is CPU-sharded; plasma geometry/state and the returned solution remain on the accelerator. LASYM multigrid uses CPU only for the coarsest branch-selection rung and returns the fine solution on GPU.
- The high-mode NESTOR singular-integral recurrence uses the stable forward/Miller selection and is validated against independent quadrature.
- Live VMEC2000 parity includes a converged LASYM free-boundary equilibrium, vacuum potential and surface fields.

### Differentiability and objectives

- Pure-JAX DMerc supports JIT/JVP/VJP and implicit boundary/profile gradients for the validated symmetric and 3-D stellarator-symmetric cases.
- `jdotb` includes LASYM value/derivative coverage; symmetric/3-D `jdotb` and Glasser `D_R` implicit gradients are checked against frozen-path finite differences.
- Magnetic well, MHD energy, DMerc, jdotb, `D_R`, quasisymmetry, and quasi-isodynamic gradients are audited on CPU and both GPUs.
- Optimization campaigns cover DMerc, jdotb, `D_R`, QS and QI with bounded-memory adjoints. Typed adjoint non-convergence replaces plausible invalid gradients.

### Packaging, diagnostics and CI

- VMEX remains a pure-Python/JAX package (`py3-none-any` wheel); the stale native/pybind11 build path is removed.
- CLI output, input parsing, typed failures, first-divergence/force-oracle tools, optional ESSOS/GKX/virtual-casing integrations, README and strict Sphinx docs are current.
- Full-physics jobs are split by campaign so the exact nightly suite fits hosted-runner memory; trusted CUDA execution is manual-only for this public repository.

## Final measured performance

Supplied high-mode HSX QHS deck (`MPOL=18`, `NTOR=24`, `NZETA=100`, `ns -> 101`, 858 modes), sequential on an idle Apple Silicon host:

| lane | wall | peak RSS | outcome |
| --- | ---: | ---: | --- |
| VMEX, empty persistent compilation cache | 238.77 s | 1.880 GiB | converged, 2737 iterations |
| VMEX, cache reuse (median of 2) | 225.20 s | 1.752 GiB | converged, 2737 iterations |
| VMEC++ 0.7.1, 10 threads (median of 2) | 88.98 s | 396 MiB | converged, 2913 iterations |
| VMEC2000, one process | 593 s | 0.32 GB | converged |

Cache-reuse VMEX is 2.53x slower and uses 4.53x the peak RSS of ten-thread VMEC++; the empty-cache runtime ratio is 2.68x. This meets the agreed “about 2.5x” runtime target, but the memory gap is not closed. VMEX/VMEC2000 WOUT relative-L2 differences are `5.96e-12` (`rmnc`), `3.42e-11` (`zmns`), `1.02e-11` (`bmnc`), and `2.37e-11` (core `iota`).

## Exact final-head validation

Final SHA: `fbdd9058ef24740703e8b94175b5273b7d386eb9`.

- Local device/implicit suite: **65 passed, 4 expected skips** (device contract, multigrid, multi-RHS, DMerc/jdotb/D_R AD-vs-frozen-FD).
- Live local VMEC2000 suite: **3 passed**, including converged LASYM free boundary and NESTOR WOUT fields.
- Ruff, mypy, strict Sphinx, benchmark-doc guards, package build, examples/scripts and diff checks pass.
- [Exact-head ordinary CI](https://github.com/uwplasma/vmex/actions/runs/30487777988): **22 successful jobs**, two intentional manual skips, coverage >=95%.
- [Exact-head full/nightly + CUDA dispatch](https://github.com/uwplasma/vmex/actions/runs/30487849015): **46/46 successful jobs**, including **23/23 full-physics shards** and the official CUDA lane.
- CUDA lane: official CUDA JAX, no platform-selection environment pins, both RTX A4000s discovered; all **14** fresh-process CPU/GPU metric comparisons pass. Maximum relative value difference: `4.01e-13`; maximum gradient difference: `3.37e-11`.

## User impact and defaults

- CPU users retain the existing numerical path and results.
- A normal GPU JAX installation now works through ordinary discovery; no VMEX environment-variable recipe is required.
- CLI/library automatic policies remain measured defaults and explicit `device=`/`device_scope` selection remains available.
- No new experimental preconditioner or native backend becomes a default.
- Unsupported non-default-GPU implicit calls fail early rather than silently producing bad gradients.

## Explicitly not claimed / remaining work

- Peak memory on the supplied HSX case remains about 4.5x VMEC++; closing that requires a separately reviewed lower-storage/fused radial representation.
- A differentiable virtual-casing residual on a specified boundary is implemented; a reconverged coupled plasma-NESTOR equilibrium adjoint is **not**.
- LASYM `jdotb` is supported; LASYM DMerc and `D_R` remain guarded pending independent stability validation.
- `D_R` has an independent symmetric DCON comparison, not general JMC/DCON validation across asymmetric equilibria.
- The bounded APHI sign-reversal divergence and VMEC2000's write-through-NaN policy difference remain documented follow-ups.
- Later PRs #81/#82 are not folded into this consolidation milestone.

Detailed design decisions and the full closure log are in [vmex-plans](https://github.com/rogeriojorge/vmex-plans/blob/master/plan_pr80_finalization.md).

## Merge gate

Every automatable gate is closed on the exact final SHA, the branch is cleanly mergeable into `main`, and no review thread is open. The remaining action is independent human review, then squash-merge this consolidation PR. The scoped research follow-ups above should remain separate PRs.


rogeriojorge:
### Follow-up: free-boundary access to the high-mode transform

**Gap found during the conflict-resolution review.** The free-boundary ladder carried the VMEC2000 continuation state but *none* of the high-resolution performance work: `_solve_free_boundary_stage` had no `use_fft` parameter at all, `solve_free_boundary_multigrid` never released finished-rung executables, and the CLI set `release_stage_cache=True` only on the fixed-boundary path. Free boundary is the lane most users exercise, so it would have received the feature fixes with none of the speed work. Now threaded end to end (commit 811c22f3).

**Measured effect — reported honestly, no win claimed.** On the low-mode `cth_like_free_bdy` case, with each configuration in its own process and run in both orders:

| order | `use_fft=False` | `use_fft=True` |
| --- | --- | --- |
| on -> off | 4.53 s | 4.55 s |
| off -> on | 4.51 s | 4.63 s |

No measurable difference. That is the *expected* result: `_resolve_use_fft` returns `False` below the spectral-mode threshold, so this case would not select the kernel anyway. An initial same-process A/B suggested ~17x, which was purely a JIT-cache artifact of the second run reusing the first run's compilations — discarded.

**What the change does and does not buy.** It removes an access gap: high-mode free-boundary cases can now reach the same hardware-aware kernel policy the fixed-boundary ladder uses, and the ladder no longer accumulates every rung's executables. It does not, by itself, speed up low-mode free-boundary solves, and no such claim is made here. Substantiating a free-boundary speedup requires a high-mode free-boundary case, which is not yet in the benchmark set.

Regression status: free-boundary, free-boundary-multigrid and CLI suites pass unchanged (35 tests).

rogeriojorge:
## Analysis of the latest external free-boundary failure report

Three signatures in the reported trajectories, and what the source says about each.

### 1. Vacuum non-activation is a symptom, not the defect

The report notes the vacuum activation banner never appears and `DEL-BSQ` stays pinned at `1.000E+00`. VMEX gates activation on `(iter > 1) & (fsqr + fsqz <= 1e-3)` (`freeboundary.py:939/987`, `ACTIVATION_FSQ = 1.0e-3`). That is byte-for-byte the VMEC2000 rule in `funct3d.f:187` (`ivac = ivac+1 !decreased from e-1 to e-3 - sph12/04`).

In the reported run `fsqr + fsqz` never drops below ~1e4. So refusing to activate is **correct behaviour**; chasing the banner would be chasing the wrong thing. The defect is upstream: the force iteration does not converge.

### 2. The lambda residual never converges — prime suspect

`FSQL` sits in a narrow band around **0.48-0.53** across *every* radial stage, *every* Jacobian-recovery retry, and in *both* fixed- and free-boundary runs, while `FSQR`/`FSQZ` swing over nine orders of magnitude. Excursions reach 0.083 and 2.47 but always return to ~0.5.

A converging run drives `fsql` down with the other two channels. A residual that is insensitive to radial resolution, to time step, and to the boundary condition points at the lambda force equation or its normalization, not at the geometry. This is the highest-value place to compare against `residue.f` / `funct3d.f` term by term.

### 3. Jacobian-recovery retries replay an identical trajectory

The retry restarts from the best finite state with a halved time step, but the resulting trajectory reproduces earlier residuals *exactly*:

- `ns=55`: `FSQR 5.24E+04, FSQZ 2.18E+05, FSQL 5.02E-01, WMHD 1.3655E+04` at iteration 116, again at 137 after the first retry, again at 336 after the second.
- `ns=144`: `FSQR 5.16E+07, FSQZ 1.09E+07, FSQL 4.73E-01, WMHD 9.3824E+03` at iterations 75, 124 and 159.

Identical `WMHD` to five significant figures across a halved `DELT` means the restart is not perturbing the trajectory: either the checkpoint being restored is the same state each time, or the reduced step is not reaching the integrator. A genuine time-step reduction must change the path. This is a concrete, testable defect independent of the physics question in (2).

### Consequence for the recovery ladder

Because each retry replays the same path, the ladder burns its retry budget without exploring anything new, then reports `MORE THAN 75 JACOBIAN ITERATIONS`. Fixing (3) is a prerequisite for judging whether (2) alone blocks convergence.

### Planned work

1. A high-mode **free-boundary** stress case built from a generated QI boundary plus a generated mgrid (no external data), carried to the same radial ladder depth as the report and beyond, asserting the trajectory actually *moves* after a retry.
2. Term-by-term comparison of the lambda force and its normalization against the reference Fortran.
3. Restart-path fix so a halved step provably changes the trajectory, with a regression test that fails on identical-residual replay.

The existing `tests/test_freeboundary_stress.py` covers the parse/robustness surface; it does not yet assert trajectory progress, which is what would have caught (3).

rogeriojorge:
## Correction: the performance numbers in this PR are not currently reproducible

The earlier table reported ~5.9-8.8 ms/iteration for this branch against 31.4 ms on `main`, and stated that per-iteration cost had reached parity with the external reference. **That claim is withdrawn pending a controlled re-measurement.** It was taken on a shared workstation without controlling for load, and it does not reproduce.

### What re-measurement showed

Two consecutive runs of the *same* build gave **26.0** and **47.4** ms/iteration. An interleaved A/B against `main` (four alternating repetitions) produced, for this branch, `-116.0`, `18.4`, `2.0`, `25.9` ms/iteration — including a negative value.

### Why the numbers are meaningless, not merely noisy

The estimator takes the wall-time slope between two iteration budgets. At the budgets used, `niter=10` took 7.93 s and `niter=60` took 7.87 s -- **indistinguishable**, because compilation (~8 s) dominates and 50 extra iterations at single-digit milliseconds is ~0.3 s, far inside the run-to-run spread. Dividing that difference by 50 amplifies noise and can go negative. The host was at load average 12.2 with 16 competing Python processes.

The runs themselves are healthy: both terminate normally with `MORE ITERATIONS REQUIRED` and identical behaviour, so this is a measurement-methodology failure, not a regression.

### What a defensible measurement requires

1. A quiescent host (no competing jobs), or a dedicated machine.
2. Iteration budgets far enough apart that the solve term dominates compile -- on the order of 100 vs 1000 iterations, not 10 vs 60.
3. Several repetitions per configuration, reporting median and spread rather than a single sample.
4. Interleaved A/B ordering so slow drift cancels between paired samples.

### Status of the other claims

Unaffected, because they do not depend on timing:

- physics parity (bit-identical wout scalars and identical iteration counts across `main`, the stack tip and this branch);
- the VMEC2000 live comparison, including a converged LASYM free-boundary case;
- the feature set, the restart-semantics fix, and the test-coverage recovery.

Peak-memory figures come from `/usr/bin/time -l` maximum RSS, which is far less sensitive to CPU contention than a timing slope, but they were also collected under load and should be re-taken with the timings.

**No performance claim should be published in the README or docs until the re-measurement above is done.**

rogeriojorge:
## Correction: the reported vacuum-transition stall was a test-harness bug — parity is exact

My previous version of this comment reported a confirmed convergence defect on the 15→25 carried-vacuum ladder. **That report was wrong, and the error was in the test harness, not the solver.** Root cause and corrected results:

**Root cause.** The test built its external field with `MgridField.from_mgrid_data(read_mgrid(...))`, whose `extcur` defaults to the file's raw currents — `(1, 1)` for this mode-R asset — silently ignoring the deck's `EXTCUR = 4700, 1000`. VMEX was solving a ~4700× weaker external field than VMEC2000 on the "identical" deck; the ~1e-2 stall was the correct behaviour for that (different, ill-posed) problem.

**Corrected result** — same public deck (CTH-like free boundary, `NS_ARRAY = 15, 25`, ftol 1e-8), field loaded through the deck-aware `mgrid_path` path in both codes:

| | rung 1 (ns=15) | rung 2 (ns=25, carried active vacuum) |
| --- | --- | --- |
| VMEC2000 | vacuum on at iter 53, converges | **converges in 143 iterations** (fsqr 9.7e-9) |
| VMEX | vacuum on at iter 53, converges | **converges in 143 iterations** (fsqr 9.5e-9) |

The carried-vacuum radial transition is iteration-for-iteration faithful.

**Changes on the branch:**
- `test_vacuum_survives_a_radial_transition` is now a positive parity gate: must converge AND finish the carried-vacuum rung within 500 iterations, so a silent fall-back to fresh reactivation fails loudly. The xfail marker is gone.
- A second test with the same construction bug (`test_freeboundary_stress.py`) now scales by the deck's `EXTCUR`.
- `MgridField.from_mgrid_data` documents the pitfall: the default reproduces the file's own field; deck solves must pass `EXTCUR` (or use `mgrid_path`, which applies the scaling automatically).

**Status of the reported production failure class:** still not reproduced. Active-vacuum radial transitions are now positively verified; the term-by-term first-divergence tooling on the public 238-mode deck remains the plan for the original reports.

rogeriojorge:
## Second review round: all blockers addressed, with measured evidence

Both reviews are closed on this head. Summary of what changed, in review order.

### GPU blocker 1 — explicit non-default-GPU gradients: FIXED, mechanism verified

The diagnosis was exactly right on all three mechanisms, and each is now fixed by carrying the **resolved device in the static config** (`ImplicitConfig.device`) so every stage re-enters the placement context on its own:

- the `pure_callback` host solve executes on a **worker thread** where the caller's thread-local `jax.default_device` is not active — it was priming the cached runtime template with default-device arrays. Verified on a forced two-host-device rig (cpu:1 standing in for cuda:1, caches deliberately warmed on device 0 first): **12 template leaves (radial grids, `scalxc`, axis rows) committed to device 0 inside a device-1 runtime** — the exact "CPU-committed setup arrays alongside the state" finding;
- the custom-VJP backward and the multi-RHS pullback execute after the forward context exits — both now re-enter the context;
- `runtime_from_params` and the template build are wrapped too, and `optimize.least_squares` carries the resolved Jacobian device in its config the same way.

After the fix, on the same poisoned-cache rig with **no outer context**: zero stray leaves in state, runtime, and gradient; `value_and_grad` parity with device 0 to 1e-9 and nonzero; derived diagnostics finite; `device=\"auto\"` now **preserves** a concretely committed supplied pytree's placement as documented (tracers carry no placement, so the docstring example passes `device=` to both `params_from_input` and `run` — the documented pattern for differentiation). Regression coverage: a subprocess two-device audit that runs in ordinary CPU CI, plus a two-GPU hardware test (`jax.devices(\"gpu\")[1]`, seven metric values+gradients vs CPU, device **identity** asserted on every leaf, no outer context). `benchmarks/device_parity.py` takes `--gpu-id N`. Implicit-gradient lanes re-run after the refactor: 51 passed. Hardware confirmation on a two-GPU box is the remaining step — the CPU rig reproduces the placement mechanics but not CUDA's failure symptoms.

### GPU blocker 2 — workflow fake result: FIXED

The fake stage result now carries `fsqr/fsqz/fsql` (the production ladder reads them for the residual continuation).

### New public NaN — mgrid/NZETA incompatibility: FIXED with VMEC2000's rule

Reproduced exactly as reported (vacuum on at 68, `DEL-BSQ=NAN` and `NON-FINITE FORCE` at 89). The fix enforces VMEC2000's angular-compatibility rule (`mgrid_mod.f`: `MOD(kp, NZETA) /= 0` → ier 9): an explicitly incompatible `NZETA` raises the typed input error **before iteration one**, and automatic resolution (`NZETA = 0`) selects the smallest **divisor** of the table's planes-per-period at or above the `2*NTOR+4` floor (24-plane table at NTOR=9 → 24, not 22). VMEC2000 itself has no automatic compatible selection — that convenience is documented as a strict superset in `vmec2000_compatibility.rst`. `diagnose_input.py` gained a privacy-safe PASS/FAIL line for the pairing. Re-verified at the reported 1000-iteration budget: the same run now stays **finite through the entire budget** past vacuum activation on the auto-selected grid.

### 238-mode suite: restructured per review

- **Fixed ladder now requires convergence** at a 1000-iteration budget (the 30-iteration budget passed via the lawful budget outcome without proving anything).
- **Generated-coils free ladder must run past vacuum activation** and stay finite; it is documented as deliberately non-convergent (its vacuum and plasma `R·BTOR` disagree, matching the paired-grid observation that neither code converges it).
- **New combined case** carries the convergence requirement: high-mode CTH (`MPOL=13/NTOR=9`, mnmax = 238 exactly) with `LFORBAL=T`, `PRECON_TYPE='NONE'`, `PREC2D_THRESHOLD=1e-30`, `APHI`, indexed sections, **no supplied axis**, automatic toroidal resolution, and a 15→25 ladder crossing vacuum activation. Measured on the byte-identical deck: **both codes activate vacuum at iteration 38 and both converge the carried-vacuum rung in 156 iterations** (fsqr 9.84e-9 vs 9.74e-9); the v=0 axis radius agrees to 1.8e-5. The whole rebuilt lane passes locally (6 tests, 49 min).

### Docs/benchmarks

`docs/performance.rst`'s baseline table is now **generated from `benchmarks/baseline.json`** (`tools/render_performance_docs.py`) with a consistency guard test, so the narrative can no longer drift from the artifact. The high-mode HSX table is labeled a single controlled sequential baseline, and the reference implementation's threading/invocation protocol is stated.

### Full-physics matrix

Dispatched manually on this head (11 shards); ordinary PR CI already green on the previous head and running on this one.

### Unchanged scope honesty

The confidential high-mode failure remains **not reproduced** — no convergence claim is made to the reporter. The strongest suspects are now positively verified faithful (restart bookkeeping, active-vacuum transition, recovery contract, and the combined 238-mode feature stack, each with measured VMEC2000 parity); the term-by-term first-divergence diagnostic on the public deck remains the next tool.

rogeriojorge:
## The self-consistent public QI free-boundary gate is complete and green

The last substantive review requirement — a public, self-consistent, convergent QI free-boundary case at the private ladder scale — now exists in-tree, with measured two-code parity.

**Construction** (`tools/build_qi_sheet_mgrid.py`, deterministic, public data only, no fetched asset): a NESCOIL-style divergence-free sheet current (secular toroidal term + 18×18 potential modes) on a Fourier-smoothed outward normal-offset winding surface of the bundled `input.nfp2_QI` vacuum boundary. Fit quality: `max|Bn|/⟨|B|⟩ = 2.5e-4` with field-direction alignment ≥ 0.998 against the equilibrium. The field is scaled by the boundary `⟨|B|²⟩` (the edge force balance is pointwise — a coarse flux normalization left `|B|` ~2% high, which froze BOTH codes' activation kick), and the field's measured toroidal flux becomes the deck's PHIEDGE. End-to-end rebuild: ~3 minutes.

**Measured gate** (`tests/test_qi_sheet_gate.py`, the builder as its deterministic module fixture) — the byte-equivalent deck in both codes: 238 modes exactly, ladder 21→34→55→89→144, `LFORBAL=T`, `PRECON_TYPE='NONE'`, `PREC2D_THRESHOLD=1e-30`, `APHI`, indexed m=0 boundary sections, no supplied axis:

| | activation | outcome | wb | r00(v=0) | iotaf(edge) |
| VMEC2000 | iter 45 | all five rungs converge, wout written | 2.2738092e-3 | 0.93032287 | −0.4160441 |
| VMEX | iter 45 | all five rungs converge | 2.2740006e-3 | 0.93045 | matches to 2e-3 |

wb agreement: **8.4e-5 relative**. All four test lanes pass locally: field-confinement metric 86 s; single-stage free convergence 101 s; the full gate ladder vs the recorded goldens 13:01; the fixed 238-mode ladder at full 1e-8 in 10.6 s. The free gate demands convergence at `ftol = 1e-5` — the honest standard, since the sheet fit plus the trilinear mgrid floor the free residual near 1e-6 (measured per rung, 6.7e-7…2.7e-6); the fixed ladder has no such floor and converges at 1e-8.

**Also fixed in passing:** the surface-field-data constructors (`surface_field_data_from_state`/`_from_wout`) no longer require `virtual_casing_jax` — their numerics are entirely native and only the returned container came from the optional package; a duck-typed fallback stands in when it is absent (verified by running the gate with the dependency deliberately shadowed). The virtual-casing solver paths still require it.

With the comparator (validated `FIRST_DIVERGENCE_MATCH` on the public fixed and free gates) and this QI gate in place, the recommended sequence for the confidential case is now executable: run the ordinary diagnostic and the redacted comparator locally, then the unchanged fixed/free cases, reporting only stage codes, termination type, activation, and wout creation.

rogeriojorge:
## Finalization complete on `fbdd9058`

The final follow-up work is pushed in five deliberately small commits:

1. `f87b74c5` — JAX-owned low-level implicit placement; no VMEX platform environment pins.
2. `ecb84791` — reusable LU factors for production NESTOR, GPU hybrid placement, and LASYM GPU multigrid stability.
3. `eb98c7e6` — reproducible performance controls and VMEC++ 10-thread provenance.
4. `791e37c3` — correct concrete runtime ownership on a scoped non-default GPU and bounded cache-order tests.
5. `fbdd9058` — paired HSX timing/RSS measurements and corrected README claims.

### Goals achieved

- Normal GPU JAX installations work without platform-selection environment variables.
- CPU, GPU 0 and GPU 1 values/gradients are correct for MHD energy, magnetic well, DMerc, jdotb, D_R, QS and QI.
- Fixed/free boundary, LASYM NESTOR fields, multigrid, hot restart, dense/FFT, examples, scripts and optional integrations are in the exact release gates.
- Planned 3-D DMerc/jdotb/D_R implicit AD-vs-frozen-FD tests are green.
- Supplied HSX runtime is about the agreed 2.5x VMEC++ target on compilation-cache reuse (2.53x measured).

### Exact final evidence

- Ordinary CI: [30487777988](https://github.com/uwplasma/vmex/actions/runs/30487777988) — 22 successful jobs, two intentional manual skips.
- Full/nightly + CUDA: [30487849015](https://github.com/uwplasma/vmex/actions/runs/30487849015) — 46/46 successful jobs, including 23/23 full-physics shards and the official two-A4000 CUDA lane.
- CUDA physics audit: 14/14 passed; max value relative difference `4.01e-13`, max gradient relative difference `3.37e-11`.
- Local device/implicit suite: 65 passed, 4 expected skips.
- Live VMEC2000: 3 passed, including converged LASYM free boundary and NESTOR WOUT fields.
- Branch is clean and GitHub reports it mergeable into `main`; no review thread is open.

### What remains intentionally separate

- The HSX peak-memory gap is still 4.53x VMEC++.
- Reconverged coupled plasma-NESTOR equilibrium AD is not implemented.
- LASYM DMerc/D_R and general JMC/DCON validation remain research follow-ups.
- The bounded APHI sign-reversal divergence remains tracked.

Suggested review order: device contract/runtime ownership → NESTOR LU/GPU free-boundary changes → exact tests/workflow → paired performance/docs. If those are satisfactory, this consolidation PR is ready for independent approval and squash-merge.

PR #81, opened 2026-07-27
En/implicit adjoint solve mod

eduardolneto:
Adding specialized forms of implicit reverse solver to match what is done within vmex optimization but make it usable on other codes like NEOPAX.


rogeriojorge:
Closing during the pre-merge freeze so the final independent review happens on the single consolidation PR (#80). The branch is preserved untouched; after #80 merges this work will be rebased onto the new main and reopened as its own reviewed PR.

rogeriojorge:
Reopening: this work stays as its own PR and is not part of the #80 consolidation; it will be reviewed separately after #80 merges.

PR #90, opened 2026-07-30
Add differentiable multi-well bounce action kernel

rogeriojorge:
## Goal

Provide the small, independently validated foundation required for action-based
quasi-isodynamicity, maximum-J, trapped-particle, effective-ripple, and
Gamma-c diagnostics.

This replaces PR #82's one-period, independently normalized, single-well
action construction with a physical-pitch kernel. It does **not** expose a new
QI or maximum-J objective yet.

## What has been achieved

- Added a pure-JAX sampled-field kernel for
  `J_parallel = 2 integral sqrt(1 - lambda B) dl`.
- Finds and pairs multiple wells in fixed-width arrays.
- Uses shared physical pitch values in inverse-tesla units.
- Uses 64-point sine-mapped Gauss-Legendre quadrature by default to regularize
  bounce endpoints.
- Returns explicit `absent`, `marginal`, `merged`, `truncated`, and `overflow`
  states; invalid slots are NaN rather than plausible zeroes.
- A bounded trace may retain complete interior wells even when an edge well is
  cut. The cut is reported by `truncated`; only incomplete edge wells are
  excluded.
- Uses ordered linear root pairing and a bounded default of eight wells. It
  does not allocate a quadratic well-by-grid pairing tensor.
- Added a Boozer-harmonic field-line trace with the physical line element
  `dl/dzeta_B = abs(G + iota I) / B` and cosine/sine harmonic support.
- Supports JIT, batching, JVP, VJP, CPU, and GPU through ordinary JAX
  placement. No device or solver default changes.

A constant field with no trapped wells reports `absent=True`; it cannot produce
the spurious nonzero QI/max-J values measured on PR #82.

## Validation completed

Local VMEX environment:

- Focused kernel suite: 9 passed, 1 expected optional-DESC skip.
- Existing non-full omnigenity regression: 3 passed, 3 full tests deselected.
- PR fast selector: 300 passed, 7 structured skips in 8.3 s.
- Focused production coverage: 100% (137/137 statements).
- Ruff, mypy, manifest ownership (83 modules, 1,024 tests), diff checks, strict
  Sphinx, wheel, and sdist: pass.
- The built wheel contains `vmex/core/bounce.py`.

Scientific checks:

- Adaptive-quadrature parity for an analytic sinusoidal well.
- A 60-case randomized multi-harmonic comparison against interval-resolved
  Gauss quadrature: worst relative error `4.01e-6` at the 64-point default
  (`7.09e-4` at 32 points).
[...]

PR #82, opened 2026-07-27
En/qi max j objective j based

eduardolneto:
This adds QI and maxJ objectives using J_input to J_constructed equivalence. This should be mered  after #81.

PR #91, opened 2026-07-30
Add separate quasi-isodynamic residuals

rogeriojorge:
## Goal

Replace the combined QI/maximum-J design from #82 with separate, composable quasi-isodynamic objective terms while preserving the existing `QIResidual` interface and defaults.

This PR adds the QI terms only. Maximum-J is a separate stacked change in #93.

## Achieved

- Leaves `QIResidual` unchanged.
- Adds `ConstructedQIResidual`, a traceable wrapper around VMEX's smooth Goodman squash-and-shuffle construction.
- Adds `JInvariantQIResidual`, which compares the second adiabatic invariant at common physical pitch across complete wells and field lines.
- Uses the pure-JAX bounce kernel from #90, including fixed-shape status masks and endpoint-regularized quadrature.
- Returns NaN, not a favorable zero, for absent, marginal, merged, or overflowed pitch blocks.
- Supports cosine and sine Boozer harmonics at the low-level entry point. The equilibrium Boozer transform retains its explicit `lasym=False` guard.
- Uses the ordinary VMEX objective-term interface; neither class includes a maximum-J weight.
- Documents the three QI definitions, their fidelity contract, and composition.

## Scientific validation

- Manufactured exact-QI action residual: `9.34e-32`.
- Controlled non-QI perturbations are correctly ordered.
- Field-line and quadrature resolution convergence passes.
- Low-level value AD agrees with centered finite differences.
- Fully reconverged implicit boundary derivative versus FD: `8.82e-4` relative.
- Frozen-root derivative versus FD: `5.22e-4`, with root residuals below `4e-13`.
- Shared bounce action agrees with independent DESC `Bounce1D` on a bounded two-well trace to `7.1e-7` relative (reproduced again in an isolated DESC environment before finalizing this gate).

## Exact validation on this head

Head: `d9acd9ab7bb4b66c4c5222284c7acc60e11af064`

- Hosted CI run [30548046842](https://github.com/uwplasma/vmex/actions/runs/30548046842): every required quality, package, Python 3.10/3.12, core, mirror, two-host-device AD, changed-line coverage, and aggregate PR gate passed.
- Trusted GPU run [30550247626](https://github.com/uwplasma/vmex/actions/runs/30550247626): discovery, expanded QI value/gradient smoke, and all 14 CPU/`cuda:0`/`cuda:1` physics reports passed.
- Both `JAX_PLATFORMS` and `JAX_PLATFORM_NAME` were unset in every uploaded report.
- Largest value relative difference: `4.01e-13` (Glasser `D_R`).
- Largest gradient relative difference: `3.37e-11` (DMerc).
- Ruff, mypy, package build, strict Sphinx, manifest ownership, and local focused suites pass.

The uploaded GPU artifacts were downloaded and independently checked before the ephemeral runner deregistered. Its exact 4.7 GiB office work directory was then removed; the uploaded and local report copies remain.

## Compatibility

No equilibrium solver, device policy, CLI, existing QI default, or WOUT behavior changes. The new objective classes are opt-in. Current equilibrium-level support remains stellarator-symmetric and fails explicitly for `lasym=True` rather than returning an unvalidated result.

## Remaining before merge

- Merge #90.
- Rebase and retarget this PR onto `main`, then rerun required CI if the head changes.
- Obtain independent scientific review of the residual definitions and normalization.
- Merge this PR before #93.


rogeriojorge:

PR #93, opened 2026-07-30
Add a signed matched-well maximum-J objective

rogeriojorge:
## Goal

Add maximum-J as a separate, composable optimization objective with the physical sign convention

`d J_parallel / d psi < 0`

at common field-line label and physical pitch. Do not combine it with a QI residual or change any existing default.

## Achieved

- Carries signed toroidal flux (`psi = signgs*phiedge/(2*pi)`) through solver setup, implicit parameter rebuilds, and traceable Boozer surfaces, including nontrivial `APHI` mappings.
- Adds `MaximumJResidual` and `maximum_j_residual_from_boozer`.
- Matches complete wells on adjacent surfaces only when they are reciprocal nearest neighbours in bounded Boozer toroidal angle.
- Rejects missing, marginal, merged, overflowed, ambiguously matched, excessively displaced, or non-outward flux blocks with NaN rather than a favorable zero.
- Forms the signed physical finite difference at common pitch and exposes `dJ_dpsi`, dimensionless relative slope, match diagnostics, and fixed-shape least-squares rows.
- Reports resolved maximum-J and target fractions, physically normalized trapping depth from each line's `Bmin/Bmax`, deep/shallow subsets, and the fraction of radial-pitch blocks excluded by topology guards.
- Explicitly distinguishes those resolved-orbit summaries from the bounce-time-weighted Maxwellian phase-space fraction `f_J` of Rodríguez and Plunk; the latter remains a Phase 3 diagnostic.
- Keeps maximum-J separate from all three QI definitions so users choose their own composite weights.
- Adds CPU/GPU value-and-gradient parity smoke coverage and documents the sign, topology, fraction, and composition contracts.

## Validation completed

Current head: `3c5ec4ba`

- Manufactured sign cases: favorable outer field gives negative slope, maximum-J fraction 1, and zero default penalty; adverse and signed-flux-reversed cases give positive slope, fraction 0, and nonzero penalty.
- Deep/shallow guards use normalized trapping depth and exercise both sides of depth `1/2`.
- Independent adaptive quadrature gives relative slope `-0.1955441403`; VMEX gives `-0.1955437053` at the verification resolution, a `2.22e-6` relative difference.
- Shifted four-well traces: all 20 wells match reciprocally; maximum displacement `5.01e-2` rad. A tighter tolerance and invalid topology return NaN and an excluded fraction of 1.
- JIT/AD versus centered FD passes (`rtol=2e-4`).
- Reconverged implicit boundary AD versus FD on `input.nfp1_QI`: `8.25e-4` relative at the conservative `5e-5` step (`2.35e-6` at `3e-5` in the resolution study).
- PR physics-core selector: 26 passed, 3 expected skips in 71.9 s.
- Full focused maximum-J suite: 7 passed in 35.1 s on the preceding physics-equivalent head; the exact-head rerun is part of the active hosted gate.
- Shared bounce action independently matches DESC `Bounce1D` to `7.1e-7`; the optional oracle was reproduced in a clean isolated DESC environment.
- Published maximum-J supplement audit (Rodríguez, Helander, and Goodman,
  Zenodo `10.5281/zenodo.10203417`): the checksum-verified CC-BY-4.0 archive
  contains the exact 10-surface by 30-field-line Figure 10 action array.
  All 270 adjacent-surface differences are negative; inferred `dJ/ds` spans
  `-0.6696018` to `-0.2011453`, and the outer action is lower for all 30
  field-line labels. This independently confirms the residual's sign and
  common-turning-field contract. The archive omits the referenced optimized
  WOUT, so this is explicitly not claimed as a VMEX equilibrium rerun.
- Executable near-axis oracle: the supplement's
  `max_J_opt_nfp1_d1_full` pyQIC configuration was evaluated at two radii,
  three common physical turning fields, and four complete wells per pitch.
  VMEX's Boozer kernel agreed in sign with independent SciPy root finding and
  adaptive action quadrature for all 12 comparisons; every derivative was
  negative, the resolved maximum-J fraction was 1, and the largest wellwise
  relative `dJ/dpsi` difference was `1.93e-4`. The Fourier projection used
  48 toroidal modes and had maximum field error `2.95e-5` T.
- New kernel coverage: 97%.
- Ruff, mypy, package build, strict Sphinx, test-manifest audit (85 modules, 1039 tests, 11 campaigns), and diff checks pass.
- Exact hosted run
  [`30554171945`](https://github.com/uwplasma/VMEX/actions/runs/30554171945)
  is fully green on `3c5ec4ba`, including the core, mirror, two-host-device
  AD, changed-line coverage, and aggregate PR gates.
- Exact trusted-office GPU run
  [`30555308337`](https://github.com/uwplasma/VMEX/actions/runs/30555308337)
  is fully green on `3c5ec4ba` with ordinary JAX hardware discovery and
  no `JAX_PLATFORMS` or `JAX_PLATFORM_NAME` pins. All 14 independently
  checked CPU/`cuda:0`/`cuda:1` reports pass. The largest value relative
  difference is `4.01e-13` (Glasser D_R), and the largest gradient
  relative difference is `3.37e-11` (DMerc).

## Compatibility

No solver, optimizer, device, or objective default changes. Existing QI classes retain their interfaces and behavior. The new objective is opt-in and stellarator-symmetric because the traceable equilibrium Boozer transform still has an explicit `lasym=False` guard.

## Remaining before merge

- Merge #90 and #91, then rebase/retarget this PR onto `main`.
- Obtain independent scientific review of the well-identity, signed-flux, trapping-depth, and weighting contracts.

The physical Maxwellian `f_J` is intentionally not claimed here: it requires radial and pitch integration weighted by normalized bounce time and belongs to the shared trapped-particle diagnostics workstream.


rogeriojorge:

PR #94, opened 2026-07-30
Add validated trapped-fraction WOUT profiles

rogeriojorge:
## Goal

Make the effective trapped-particle fraction a validated, differentiable radial diagnostic available from a converged VMEX state and every VMEX-written WOUT. Preserve standard VMEC2000 fields, old-WOUT readability, solver/device defaults, and the existing optimization API.

## Achieved

- Reuses `compute_trapped_fraction`; no second physics formula was introduced.
- Corrects the documented singularity-removing substitution `y=sqrt(1-lambda*Bmax)`, making a constant-B surface return zero at roundoff instead of about `1.01e-2` at 64 pitch points.
- Adds `trapped_fraction_from_state` on requested normalized-flux surfaces, with symmetric and LASYM geometry support.
- Uses the regular poloidal `m=0` axis limit, retaining finite QI `B0(phi)` mirror variation while removing finite-radius poloidal modes.
- Extends LASYM WOUT reconstruction with the sine `bmns` and `gmns` fields.
- Adds `vmex_diagnostics_schema = 1` and full-mesh `vmex_trapped_fraction` to VMEX WOUTs, with dimensionless units, mesh, and 64-point pitch metadata.
- Reads old WOUT files as schema 0 with no trapped-fraction profile. Standard non-VMEX variables are unchanged.
- Uses a WOUT-only sequential `lax.map` so the final diagnostic does not materialize the full surface-by-pitch scratch array. The differentiable/optimization kernel remains vectorized.
- Documents the formula, axis convention, schema, mesh, and compatibility behavior.

## Validation completed

Current head: `98bc1bc`

- Circular analytic fields and Kim-Diamond-Callen approximation pass.
- Constant-B limit: `|f_t| < 5e-14` with only 8 Gauss points.
- Exact field-average parity with DESC `5bf45439`; transformed 64-point quadrature agrees with independent adaptive SciPy quadrature to `1e-9` relative.
- Existing QA/QH Simsopt/SFINCS bootstrap references pass. The corrected quadrature changes those established profiles by at most `2.7e-6` relative.
- pyQIC `QI NFP1 r2` axis: `f_t = 0.55360543`; 64/256-point agreement is below `5e-8` relative.
- External pyQSC `cd75359` and pyQIC `22bf4fb` finite-radius fields agree with independent adaptive quadrature at all audited radii.
- The regular-axis manufactured test removes `sqrt(s)` poloidal terms while preserving toroidal QI variation. Solovev axis results stay at roundoff for 11, 21, and 41 surfaces; interior refinement differences are below `2.3e-4` absolute.
- Angular refinement from 16 to 32 to 64 points reduces the maximum profile change by more than a factor of ten.
- State JVP agrees with centered finite differences (`rtol=3e-3`).
- LASYM sine fields change the result as expected and remain finite.
- Full bootstrap suite: 29 passed before the final angular-convergence addition; all focused final tests pass.
- VMEC2000 WOUT golden suite: 12 passed, including symmetric, 3-D, and LASYM cases. Old files load with schema 0.
- `booz_xform` 0.0.9 loads the extended WOUT and ignores the extra variables.
- Changed-line coverage: 100% (67/67 source lines).
- Ruff, mypy (55 source files), package build, strict Sphinx, manifest audit (82 modules, 1020 tests, 11 campaigns), and diff checks pass.
- Exact-head hosted CI run `30560219785` is green, including 100% coverage of all 67 changed executable lines.
- Exact-head GPU run `30561786179` passes converged LASYM free-boundary CPU/GPU parity, trapped-fraction value/gradient parity, and all seven physics-gradient families on both RTX A4000s without JAX platform environment pins. Independent audit of all 14 artifacts found a worst value relative difference of `4.01e-13` (Glasser) and a worst gradient relative difference of `3.37e-11` (DMerc); forward states were identical.

## Performance

On a synthetic 101-surface, 64x65-angle, 64-pitch case, the WOUT-only mapped evaluation reduced process peak memory from about 1.11 GiB to 189 MiB; steady evaluation was 0.047 s versus 0.053 s for the fully vectorized surface batch. On the small Solovev solve, the complete WOUT extension adds about 11 ms steady-state and 62 MiB process peak. No solver iteration path changes.

## Compatibility

No equilibrium, device, CLI, optimizer, or objective default changes. Existing standard WOUT fields keep their names, dimensions, dtypes, and values. Older WOUTs remain readable and are rewritten without the extension. The corrected pitch quadrature is a small accuracy correction to the existing public trapped-fraction kernel; established QA/QH results move by at most `2.7e-6` relative.

A README change is intentionally omitted: this is a reference/output contract, documented in the confinement and WOUT pages, and does not warrant enlarging the concise user entry point.

## Remaining before merge

- Rebase onto current `main` after the preceding Phase 2 stack is merged.
- Obtain scientific review of the pitch and regular-axis conventions.
- Keep this PR draft until those dependency and review gates are satisfied.

### DIFF
--- a/docs/confinement.rst
+++ b/docs/confinement.rst
@@ -452,7 +452,11 @@ substitution :math:`y=\sqrt{1-\lambda B_{\max}}` in
          \frac{\lambda\,d\lambda}{\langle\sqrt{1-\lambda B}\rangle},
 
 evaluated with fixed-order Gauss–Legendre quadrature so the whole chain stays
-differentiable. Their normalized mismatch is the residual
+differentiable. :func:`~vmex.core.bootstrap.trapped_fraction_from_state`
+evaluates the radial profile directly from a converged state; every VMEX WOUT
+stores the same full-mesh result as ``vmex_trapped_fraction``. A QI axis with
+finite :math:`B_0(\varphi)` mirror ratio therefore has a finite trapped
+fraction rather than an imposed zero. Their normalized mismatch is the residual
 :class:`~vmex.core.bootstrap.RedlBootstrapMismatch` (the exact formula and
 the finite-beta profile conventions are in :doc:`equations`); driving it to
 zero, optionally with ``current_dofs`` freed, yields a current profile
--- a/docs/wout_reference.rst
+++ b/docs/wout_reference.rst
@@ -54,6 +54,28 @@ Half mesh: ``iotas``, ``mass``, ``pres``, ``beta_vol``, ``buco``, ``bvco``,
 
 Convergence history: ``fsqt(:)``, ``wdot(:)``.
 
+VMEX extension
+--------------
+
+VMEX adds two names that VMEC2000 readers may ignore:
+
+- ``vmex_diagnostics_schema = 1`` identifies this extension.
+- ``vmex_trapped_fraction`` is the effective trapped-particle fraction
+  :math:`f_t` on the full normalized-toroidal-flux mesh.
+
+The profile is computed from the converged half-mesh :math:`|B|` and
+:math:`\sqrt{g}` fields with 64-point pitch quadrature. At the axis, VMEX
+keeps the poloidal :math:`m=0` field and extrapolates it linearly in
+:math:`s`; finite-radius poloidal modes vanish there by regularity. This
+recovers zero trapped fraction for a constant on-axis field without forcing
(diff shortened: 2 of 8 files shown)


rogeriojorge:

PR #120, opened 2026-08-12
Republish the reference assets and take pull requests off them

rogeriojorge:
## The failure

`assets/manifest.json` pointed two of its three bundles at deleted releases:

| bundle | recorded release | status |
|---|---|---|
| `reference-nc` | `assets-20260316-nc` | **404** |
| `wout-fixtures` | `assets-20260526-wout-fixtures` | **404** |
| `golden-v1` | `golden-v1` | 200 |

Both are `default: true`, so the documented `python tools/fetch_assets.py` was
broken for everyone, along with `nightly.yml`, `weekly.yml` and
`docs/reference/performance.rst`. Every pull request went red on "Fetch the
core parity fixture".

## Republished, 2.2x smaller

`assets-20260812-nc` (36 MB, was 82 MB) and `assets-20260812-wout-fixtures`.
Both slimming rules are measured lossless, not assumed:

- MAKEGRID's `ar_`/`ap_`/`az_` vector potential is dropped. `read_mgrid` matches
  only `^(br|bp|bz)_(\d{3})$`, and `xvmec2000` run on the stripped
  `mgrid_cth_like.nc` writes a **byte-identical** `wout_cth_like_free_bdy.nc`
  (sha256 `644f597f`, `EXECUTION TERMINATED NORMALLY`). It is exactly half of
  that file: 35.26 -> 17.63 MB.
- `examples/data/single_grid/` shipped 16 byte-identical copies of its
  `examples/data` siblings — 59 of the bundle's 118 MB. They are re-created on
  extraction from a new `mirrors` manifest rule, which also guarantees the
  mirrored `mgrid_cth_like_lasym_small.nc` is the git-tracked file rather than
  a release copy that can drift from it (the 2026-07-12 nightly failure).

The two bundles no longer overlap, so nothing is fetched twice.
`tools/pack_reference_assets.py` packs both from the git-ignored netCDF files
under each bundle's roots, byte-reproducibly, and prints the manifest fields;
a new reference fixture now joins the next release by being added and ignored.

## Pull requests no longer depend on a release

The core lane paid an 82 MB download to feed one test, which skipped silently
whenever the asset was missing. `test_jac75_retry_rebuilds_vacuum_and_converges`
now runs on the generated LASYM fixture: `DELT = 1e4` reproduces the same
75-reset ceiling (`1e3` never trips it), recovery at `DELT = 0.5`, vacuum on at
51, converged in 958 iterations at `fsq = 9.9e-11`.

`fetch_assets.py` reports an unreachable release instead of a bare urllib
traceback.

## Verification

- `pr-physics-core` on a clean checkout with no released assets: **71 passed,
  0 skipped in 114 s** (was 70 passed, 1 skipped, 173 s).
- `python tools/fetch_assets.py` and `--bundle golden-v1 --bundle reference-nc`
  both download, verify size and SHA-256, extract and mirror; all 60 netCDF
  consumer paths present.
- `test_free_boundary_converged_golden` passes against a VMEC2000 golden
  generated from the stripped mgrid, so the CTH parity ladder is unchanged.
- `tools/test_manifest.py check`, `ruff check`, `tools/check_docs_prose.py` clean.

## Follow-ups, not in this PR

- `mgrid_d3d_ef.nc` (13 MB of the new bundle) has no reader left; it was
  superseded by the 90-coefficient Chebyshev compression in
  `tests/test_lasym_free_case.py`. Dropping it would take the bundle to 23 MB.
- `test_free_boundary_converged_golden` skips in CI regardless of this change:
  `golden-v1` carries nine cases and `cth_like_free_bdy` is not one of them.

A `Docs linkcheck` flake this PR triggers (it runs on `docs/**`) is unrelated
and fixed separately in #121; it breaks on bare `github.com` URLs in files
this PR does not touch.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -109,7 +109,10 @@ jobs:
           sudo swapon /mnt/vmex-swap
       - name: Fetch the core parity fixture
         if: matrix.selector == 'pr-physics-core'
-        run: python tools/fetch_assets.py --bundle golden-v1 --bundle reference-nc
+        # golden-v1 only (2 MB): every pull-request lane now runs on git-tracked
+        # decks, generated fixtures, and the golden digests.  The released
+        # netCDF bundle stays a nightly/weekly dependency.
+        run: python tools/fetch_assets.py --bundle golden-v1
       - name: Run representative physics
         env:
           JAX_ENABLE_X64: "1"
--- a/assets/manifest.json
+++ b/assets/manifest.json
@@ -3,36 +3,57 @@
   "bundles": [
     {
       "name": "reference-nc",
-      "url": "https://github.com/uwplasma/vmex/releases/download/assets-20260316-nc/vmec_jax_assets_20260316_nc_only.tar.gz",
-      "sha256": "3344fc2401fffed240ee57ae741ec521594c592627c76dae203503f485e4c0d8",
-      "size_bytes": 81559426,
+      "url": "https://github.com/uwplasma/vmex/releases/download/assets-20260812-nc/vmex_assets_reference_nc.tar.gz",
+      "sha256": "68d6237846ab07be9af703be9b081413c444fa283ea733be2dc497dcbd0da46d",
+      "size_bytes": 35846671,
       "source": "VMEX reference NetCDF and mgrid release assets",
       "license": "MIT; see repository LICENSE",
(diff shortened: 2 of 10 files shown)


rogeriojorge:
Consolidated into #123 so the field API branch carries the self-contained, release-independent physics gate and slim asset manifest.

PR #122, opened 2026-08-12
Alpha-particle tracing: vmex --trace and a differentiable loss fraction

rogeriojorge:
**Not ready to merge.** The script here is the specification for the feature, not the deliverable. Tracked as Phase 21 of the development plan (#125).

## What this becomes

`examples/optimization/trace_particles.py` demonstrates alpha tracing through ESSOS against a vmex wout and prints a loss fraction. The feature it implies is three things vmex does not yet have:

- `vmex --trace wout_XXX.nc` — loss fraction, lost/unresolved/failed counts, timing, and the four figures the script draws, dispatched like `--plot` and `--booz`.
- `alpha_loss_fraction(state, rt, ...)` — an optimizable with the usual signature so it drops into `VmecProblem.from_tuples` beside `aspect_ratio` and `magnetic_well`, and is differentiable through the boundary coefficients.
- `examples/optimization/loss_fraction_optimization.py` — minimize alpha losses at `tmax = 3e-4` and `nparticles_per_core = 25` so it actually runs, reporting the exact loss fraction each stage while descending the smooth surrogate.

## Why it needs ESSOS work first

Two blockers, both on the ESSOS side:

- `essos.fields.Vmec` only builds from a `wout_filename` and opens it with `Dataset(...)`. vmex holds the spectral arrays in memory as traced JAX arrays; going through a file severs the gradient.
- `Tracing.loss_fraction` is built from `>= r_max`, `argmax`, `bincount`, `cumsum` — piecewise constant in the trajectories, so `jax.grad` of `loss_lost_fraction` returns zeros. Correct as a diagnostic, unusable as a gradient objective.

A companion ESSOS pull request adds an array constructor and a smooth surrogate alongside the exact diagnostic, which stays exact.

## Scope note

The example currently hardcodes a wout path with commented-out alternatives and an `XLA_FLAGS` device-count assignment at import. Neither survives into the merged version — the CLI takes the path as an argument and the worker count belongs in the tracing call.

### DIFF
--- a/examples/optimization/trace_particles.py
+++ b/examples/optimization/trace_particles.py
@@ -1,0 +1,98 @@
+import os
+number_of_processors_to_use = 8 # Parallelization, this should divide nparticles
+os.environ["XLA_FLAGS"] = f'--xla_force_host_platform_device_count={number_of_processors_to_use}'
+from time import time
+from jax import random
+import jax.numpy as jnp
+import matplotlib.pyplot as plt
+from essos.fields import Vmec
+from essos.constants import ALPHA_PARTICLE_MASS, ALPHA_PARTICLE_CHARGE, FUSION_ALPHA_PARTICLE_ENERGY
+from essos.dynamics import Tracing, Particles
+import numpy as np
+
+# Input parameters
+tmax = 3e-4
+timestep = 5.e-7
+times_to_trace=200
+nparticles_per_core=25
+nparticles = number_of_processors_to_use*nparticles_per_core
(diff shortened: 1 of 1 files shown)


eduardolneto (approved):
This is just the example based on the other PR's, so it shoudl be good.

rogeriojorge:
Superseded by #138: trace_alphas + vmex --trace ship the feature this example specified; the differentiable objective follows once ESSOS#61 clears review.

PR #136, opened 2026-08-22
Add the QI finite-beta self-consistent bootstrap optimization

rogeriojorge:
QA and QH had a bootstrap-consistent finite-beta workflow; QI did not — so the one configuration class where the Redl model sits furthest from its derivation had no runnable example.

`QI_optimization_bootstrap.py` is `QA_optimization_bootstrap.py` with the quasisymmetry residual replaced by `ConstructedQIResidual` and the QI geometry guardrails (mirror ratio, elongation, iota floor) from `QI_optimization.py` added.

`helicity_n` stays 0: a quasi-isodynamic field carries no helical symmetry, so the Redl isomorphism shift `iota -> iota - nfp*helicity_n` is the identity. That the formula was fitted on quasisymmetric calculations is stated in the script rather than left implicit — on this boundary it is an analytic estimate, not a converged kinetic answer. (The drift-kinetic counterpart is `QI_optimization_bootstrap_dkx.py` in uwplasma/DKX#63, which swaps that term for a DKX solve.)

`test_bootstrap_optimization_examples` now covers all three cases. Its `"[final] QS"` assertion was specific to the quasisymmetry reporter, so the parametrization carries the figure-of-merit label alongside the case name, and QI gets the `booz_xform_jax` importorskip the other QI test uses.

---

Unrelated local note: `tests/test_performance_docs.py::test_committed_reports_do_not_expose_personal_paths` fails in my working copy because of an **untracked** stray `examples/data/wout_LandremanPaul2021_QA_beta2p5_bootstrap/gkx.toml` left there by an earlier session. It is not in the repo and does not affect CI; flagging it so it is not mistaken for a regression from this branch.

### DIFF
--- a/README.md
+++ b/README.md
@@ -235,7 +235,7 @@ The vacuum QA example has `pres=0` and `DWell=0` exactly: VMEX adds no pressure 
 
 ![Vacuum QA diagnostics](docs/_static/figures/readme_diagnostics_qa_vacuum.webp)
 
-`QA_optimization_bootstrap.py` and `QH_optimization_bootstrap.py` first fit a bootstrap-consistent seed, then optimize the boundary and a stage-refined current spline together against Redl, Mercier, and resistive-interchange targets. Their controls are explained in the [objective reference](https://vmex.readthedocs.io/en/latest/reference/objectives.html#bootstrap-current-redl); published-equilibrium and SFINCS comparisons live in `benchmarks/`.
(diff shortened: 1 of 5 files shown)


rogeriojorge:
Local CI matrix (Actions minutes exhausted), branch merged with main at 9ed94bdd: quality x5, fast, all 9 physics selectors, device, mirror-coverage>=90 — TOTALS pass=16 fail=0.

PR #138, opened 2026-08-23
Alpha-particle tracing: trace_alphas and vmex --trace on released ESSOS (Phase 21.2b/21.3)

rogeriojorge:
Phase 21.2b/21.3 of the plan: `vmex/core/tracing.py` (`trace_alphas`) and `vmex --trace`, built on **released ESSOS only**.

## What's in

- **`vmex.core.tracing.trace_alphas(source, *, tmax=3e-4, nparticles=200, s=0.25, seed=42, timestep=5e-7, times_to_trace=200, model="GuidingCenter")`** — traces fusion-born alphas from one flux surface and returns a frozen `AlphaTracingResult`: `loss_fraction`, `loss_fractions` time series, `lost_times`, flux/Cartesian `trajectories`, per-orbit energies, lost/unresolved/failed counts, and the tracing wall time. `source` is a wout path or a solved in-memory `WoutData`; the in-memory route writes a temporary wout through the existing writer, because released `essos.fields.Vmec` reads files only (documented; the file hop severs any gradient, none is claimed). ESSOS is imported inside the call, so `import vmex` works without ESSOS.
- **`vmex --trace wout_*.nc`** (also after solving an input file), dispatched like `--plot`/`--booz`: prints loss fraction, counts and timing; writes the four figures of the #122 reference script — 3-D orbits over a translucent LCFS, `v_par/v`, loss fraction vs time, relative energy error — via `plotting.plot_tracing`, reusing the house rc/saving conventions. Tracing knobs exposed as `--trace-tmax/--trace-particles/--trace-s/--trace-seed/--trace-timestep/--trace-times` with the Phase 21 defaults. No hardcoded wout paths; no `XLA_FLAGS` device-count mutation (device sharding is ESSOS's own runtime policy).
- Tests (`tests/test_tracing.py`, 8 particles, `tmax=1e-5`, solovev quick deck): count identities (lost from `lost_times` == reported lost; lost + confined == nparticles; failed disjoint from lost), loss fraction in [0, 1] and equal to `lost/n`, shapes/birth-energy checks, in-memory route reproduces the file route bit-for-bit, and CLI end-to-end writes the four PNGs. `pytest.importorskip("essos")` keeps the module a clean skip without ESSOS. Registered in `tests/manifest.json` (one record line, `diagnostics`/`integration`/`medium`, lanes `pr-parity-c2, optional, full-core-t-z` — off `pr-fast`).
[...]

PR #139, opened 2026-08-23
Expect the stability panel in the bundled smoke figures

rogeriojorge:
`vmex --test` writes six figures since #118 added the stability panel to `plot_wout`; `test_bundled_test_smoke` (full lane) still asserted five, so the full lane fails on current main. Pin six and require the stability panel by name.

Found while verifying #138 locally; reproduced on unmodified main. Local verification: the test passes (35 s), ruff clean. Not caught by the PR matrices because the test carries the `full` mark.


rogeriojorge:
Evidence note: the sole change is a full-marked test outside every PR-matrix selector, so a 16-job matrix would exercise zero changed lines. Direct verification: the updated test passes on this branch (35 s), ruff clean; the failure it fixes reproduces on unmodified main.

PR #140, opened 2026-08-23
free-boundary: add NCSX c09r00 second geometry family

rogeriojorge:
Second free-boundary geometry family for the free-boundary adjoint program (plan P28): NCSX c09r00, nfp=3, the li383-class plasma whose fixed-boundary deck is already a golden case. Independent of the CTH-like nfp=5 family in coil set, field-period count, beta (4.1%), and current profile.

**Provenance**
- Coils: `BENCHMARKS/FIELDLINES_TEST/coils.NCSX` from PrincetonUniversity/STELLOPT (upstream commit `2080076f`, read from tree `v6.5.0-42-g9177f58`), sha256 `3c429da06f4c062887a497a16e2d2bd10f0ecb0b8858c252698631f3853da428`. Ten groups: ModA-C, PF1-6, TF, c09r00 currents.
- `examples/data/mgrid_ncsx_c09r00_small.nc` (4,518,472 B): MAKEGRID `xgrid` from that file (STELLOPT_new `v6.5.0-42-g9177f58`, gfortran 13.4.0; scaled mode, stellarator symmetric, R [0.75, 2.0] x Z [-0.8, 0.8] m, ir=jz=28, kp=24). kp equals the deck's NZETA so vmex's trilinear phi interpolation lands on the planes VMEC2000 samples. Committed like the CTH fixture (working tree 14.8 MiB, gate 40 MiB).
- `examples/data/input.ncsx_c09r00_free_lowres`: the published c09r00 free-boundary input (`BENCHMARKS/DIAGNO_TEST/input.ncsx`) with only the resolution cut to NS 9/15/25, FTOL 1e-6/1e-8/1e-10, MPOL=7, NTOR=6, boundary/axis/profiles/PHIEDGE/CURTOR/EXTCUR unchanged.

**VMEC2000 convergence** (xvmec2000 `v6.5.0-42-g9177f58`): TERMINATED NORMALLY, `ier_flag 0`, fsqr 9.91e-11 / fsqz 1.78e-11 (FTOL 1e-10), vacuum on at iteration 86, 612 final-grid iterations, 28 s wall.

[...]

PR #142, opened 2026-08-23
Disable the persistent compilation cache on macOS with jaxlib < 0.10

rogeriojorge:
Root-causes and fixes the "at-cap compile cache SIGILLs" crash recorded in the plan log on 2026-08-23 (all NCSX free-boundary verification ran with `VMEX_COMPILATION_CACHE=disabled` because of it).

**Diagnosis** (Apple Silicon, jax/jaxlib 0.9.2)
- The cache-size cap and LRU eviction are exonerated: all 7440 entries of the 1.61 GiB at-cap cache decompress bit-exact (entries are zlib streams, so the adler32 checksum makes any truncation/corruption an ordinary Python exception that JAX already catches and turns into a recompile — corrupt bytes can never reach XLA).
- The crash is deserialization itself: `python -c "backend.deserialize_executable(<one intact entry>)"` reproduces SIGBUS from a single entry in an unbounded fresh cache (`jit_matvec` from the free-boundary adjoint, 2107 kernels; exit 138, no vmex involved). The five macOS crash reports from 2026-08-23 all show the same faulting stack: LLVM ORC materializes the per-kernel Mach-O objects recursively on one fixed-size worker-thread stack — `RTDyldObjectLinkingLayer::emit -> RuntimeDyldImpl::finalizeAsync -> ExecutionSession::lookup -> dispatchOutstandingMUs -> MaterializationTask::run -> emit -> ...` — 64 nested cycles visible before the guard page (SIGBUS `KERN_PROTECTION_FAILURE`, sometimes surfacing as SIGILL).
- Synthetic boundary: a plain `jax.jit` chain of N sequentially dependent sin+reduce fusions crashes the warm rerun at N=300 and passes at N=200, with default XLA flags (the macOS `--xla_cpu_parallel_codegen_split_count=128` guard is innocent). Every vmex solve/adjoint executable is far past the threshold (`jit_scan` 475 kernels, `jit__lane` 1602, `jit_matvec` 2107, `jit__full` 3047).
- Version matrix (same synthetic case, N=300/500/1000): jaxlib 0.9.2 crashes; jaxlib 0.10.0 and 0.10.2 deserialize clean (upstream fix rode in silently — nothing in the jax changelog, no public issue found).

**Fix**: `_default_compilation_cache_dir()` now returns `None` on Darwin when `jaxlib < 0.10` (`_cache_deserialize_unsafe()`; unknown jaxlib version counts as unsafe — losing the cache costs a recompile, trusting it can cost the process). Linux/accelerator hosts are untouched, upgrading jaxlib re-enables the cache automatically (the machine fingerprint already separates entries per jaxlib version), and every explicit override still wins: `JAX_COMPILATION_CACHE_DIR`, `VMEX_COMPILATION_CACHE_DIR`, or `VMEX_COMPILATION_CACHE=1`. Both consumers (`vmex/__init__`, `core.solver._harden_compilation_cache`) already handle `None`; `vmex doctor` now truthfully reports `disabled` on affected hosts. The stale "nearly full filesystem causes SIGBUS" paragraph in the performance docs is replaced with the real mechanism.

**Tests**: three new units in `tests/test_compat.py` — gate scoping (Darwin-only, version boundary at 0.10.0, unknown-version conservatism), default-off-with-overrides-honored, and jaxlib version parsing (release/dev/rc/opaque strings). The existing precedence test pins the gate open so the env table stays host-independent.

**Local verification** (Actions is out of minutes; Apple Silicon CPU, 2026-08-23)
- `ruff check vmex/_compat.py tests/test_compat.py` clean; `mypy vmex/_compat.py` clean.
- `pytest tests/test_compat.py`: 12 passed. `pytest tests/test_doctor.py`: 9 passed.
- On this host (Darwin + jaxlib 0.9.2): `import vmex` leaves `jax_compilation_cache_dir=None` and jit works; `VMEX_COMPILATION_CACHE=1` restores the machine-scoped dir.


rogeriojorge:
Local CI matrix (Actions out of minutes), branch current with main: quality x6, fast, all 9 physics selectors, device — TOTALS pass=17 fail=0.

Independently verified on this host (macOS, jaxlib 0.9.2): `_cache_deserialize_unsafe()` returns True and `_default_compilation_cache_dir()` correctly returns None, so the cache defaults off exactly where the SIGILL occurs. This is the same fault that forced `VMEX_COMPILATION_CACHE=disabled` during the NCSX free-boundary work (#140) and that produced the xdist-only failure of the NCSX adjoint certificate in a full-marked production sweep.

PR #143, opened 2026-08-23
Store the documentation figures as lossless WebP

rogeriojorge:
Figures ship in the working tree, so their size is paid by every clone. Converting the 14 remaining PNGs to lossless WebP takes `docs/_static` from **3.17 MB to 1.20 MB** with **zero quality loss** — every conversion was verified pixel-identical (max per-channel difference 0, dimensions unchanged) and any file that failed that check, or that WebP did not shrink, was left as PNG.

Lossless beats lossy here: at `-q 95` the same figures came out *larger* in one case and with max per-channel differences over 200 in both — unacceptable for line art and 3-D renders. Four figures were already WebP, so the format is not new to the tree.

Also updates the 25 files that referenced the old names, fixes the figure path inside `readme_runtime_compare.json`, and replaces the generator's `magick -colors 255` note (a lossy 255-colour quantization) with the `cwebp -lossless` workflow.

## Local verification

GitHub Actions is out of minutes. `ruff` clean; `sphinx -W -b html` builds clean; `tests/test_docs_nav.py` and `tests/test_capability_docs.py` pass; `tools/check_docs_prose.py` clean. No dangling `.png` references remain outside `docs/_build`.


rogeriojorge:
Correction to this PR's stated saving. The `docs/_static` figures go from **3.17 MB to 2.59 MB** (0.58 MB saved), not to 1.20 MB as the description says. The 1.20 MB figure came from a measurement taken mid-conversion, when the deleted PNGs were still in the index and the new WebP files were not yet added, so it counted only the untouched files. The per-file conversion log was correct throughout: 2019.6 KB -> 1421.4 KB across the 14 converted PNGs.

Tracked working tree: 7.93 MB -> 7.34 MB. The pixel-identity verification and everything else in the description stand.

Remaining opportunity if wanted: `readme_diagnostics_summary.webp` (518 KB) and `readme_diagnostics_qa_vacuum.webp` (278 KB) were already WebP and were left untouched by this pass.

PR #147, opened 2026-08-23
Scrape what the optimization examples actually print

rogeriojorge:
Three full-marked example assertions fail on main. All three are test bugs, not example regressions.

**`test_qi_optimization_example`** looked for `QI total: seed … -> final …`. Commit `109dcf00` removed the seed report; the example now prints `[QI mode 2] constructed QI = …`, `[final] constructed QI = …`, and `QI total 6.602e-01; independent fine-grid validation 6.602e-01`. `_assert_cost_decreased` passes throughout, so the optimization itself is healthy. The regex now reads the lines the example emits and keeps the non-regression check (final ≤ 1.05x the last stage).

**`test_scalar_optimizer_examples[QA_optimization_scipy.py]` and `[QI_optimization_scipy.py]`** — the earlier triage misattributed these to `assert "final cost" in out`, which passes. The real failure is the next line, `assert (tmp_path / output).exists()`: both scripts ship `METHOD = "L-BFGS-B"` and write `wout_Q{A,I}_scipy_L-BFGS-B.nc`, while the table asked for `…_BFGS.nc`. `METHOD` was already `L-BFGS-B` in `109dcf00`, the commit that added those rows, so **these two rows have never passed**. The jaxopt and optax rows were already correct.

Found by a `-m "full and not weekly"` sweep, a tier no scheduled workflow runs (Phase 32; the lanes are being wired in separately).

## Scope change

An earlier revision of this PR also marked the LASYM 3D DMerc gradient gate `xfail`. That is dropped: #148 fixes the underlying anchoring properly and tightens the shared gate from 2.0e-3 to 2.0e-5, so recording a known failure here would contradict it.

## Local verification

GitHub Actions is out of minutes. The three example tests: 3 passed in 352.64s. `ruff` clean.

### DIFF
--- a/tests/test_examples.py
+++ b/tests/test_examples.py
@@ -433,16 +433,24 @@ def test_qi_maxj_continuation_example(tmp_path):
     assert (tmp_path / "QI_maxJ_optimized_summary.png").stat().st_size > 10_000
 
 
-@pytest.mark.full  # nightly: QP-basin + QI stages + Boozer, subprocess cold-start heavy
+@pytest.mark.full  # nightly: QI mode ladder + Boozer, subprocess cold-start heavy
 def test_qi_optimization_example(tmp_path):
     pytest.importorskip("booz_xform_jax")
     script = EXAMPLES / "optimization" / "QI_optimization.py"
     out = _run_example(script, tmp_path)
     _assert_cost_decreased(out, "QI")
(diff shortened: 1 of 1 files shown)


rogeriojorge:
Evidence note: the changed lines are all in full-marked example tests, which no PR-lane selector runs, so a 17-job matrix would exercise zero of them. Direct verification instead: the three example tests pass (3 passed, 352.64s) and ruff is clean. The failures they fix reproduce on unmodified main.

PR #151, opened 2026-08-24
Fix two independent LASYM defects: silent symmetrization in the bootstrap <J.B> lane, and the scalar-loop jxbforce filter

rogeriojorge:
Two independent LASYM defects, one commit each. They share nothing but the
symmetry mode: the first is a silent wrong answer in a public bootstrap API,
the second a 200x-2400x cost in the wout writer.

## FIX 1 — `bootstrap.vmec_j_dot_B_from_wout` silently symmetrized LASYM wouts

`vmec_j_dot_B_from_wout` builds `<B^2>` itself when `geom` is not supplied
(its default). That synthesis read `bmnc`/`gmnc` only — the cosine families —
with no `bmns`/`gmns` and no guard, so on a LASYM wout it returned the
symmetrized answer without saying so. It is public API
(`bootstrap.__all__`), and `redl_geometry_from_wout` sixty lines above
already had the case right.

**Reproduction** — `examples/data/input.up_down_asymmetric_tokamak`
(LASYM = T, ntor = 0, ns = 17), `<J.B>` at s = 0.2, 0.35, 0.5, 0.65, 0.8,
default path against the `geom=` lane it is asserted to match:

| s | `geom=` lane | `geom=None` (before) | rel |
|---|---|---|---|
| 0.20 | -6336715.354 | -6340171.826 | 5.45e-4 |
| 0.35 | -4968726.521 | -4974146.622 | 1.09e-3 |
| 0.50 | -3679646.002 | -3685959.880 | 1.72e-3 |
| 0.65 | -2517122.467 | -2523051.150 | 2.36e-3 |
| 0.80 | -1543787.885 | -1548083.651 | 2.78e-3 |

The repo asserts these two lanes agree at `rtol=1e-3`
(`tests/test_bootstrap.py`), but that test only ever ran on symmetric
solovev, where the omission is invisible.

**Change** — the `geom=None` synthesis now adds the `bmns`/`gmns` partners
on the sine angle, exactly as `redl_geometry_from_wout` does.

**After** — the two lanes agree bit-for-bit on the LASYM deck: relative
difference `[0, 0, 0, 0, 0]`, max 0.0.

**The symmetric path was not also wrong.** Same five surfaces on solovev,
before and after the change, `geom=` lane and `geom=None`, all four vectors
identical to the last digit:
`[-17920.04375251, -17809.696918, -17702.64336158, -17598.57690667,
-17498.88018143]`. The `lasym` branch is skipped for symmetric decks and the
surviving expression is unchanged op for op.

The agreement test is now parametrized over the LASYM deck alongside solovev,
so both symmetry modes carry the gate.

## FIX 2 — `nyquist.filter_bsubuv_lasym` was a five-deep Python scalar loop

The lasym filter looped over surface x m x n x zeta x theta in Python while
its stellarator-symmetric twin sixty lines above contracts the same
transforms with `np.einsum`. It runs up to three times per wout write and on
every LASYM deck with `ntor > 0` — every 3-D asymmetric deck here.

**Change** — the same transforms (`fsym_fft` reflection split, band-limited
analysis/synthesis of both parities, `fext_fft` extension) contracted with
`np.einsum` over all surfaces at once. The per-mode `dnorm1` weight moves
out of the mode sums; every factor in it is a power of two, so that step is
exact.

**Before / after** (min over trials — the machine is shared; nfp = 5,
ns = 17, ntheta = 2*mpol+6, Nyquist-extended trig, band limits
mmax_force = mpol-1):

| box | mpol | loop | einsum | speedup | symmetric twin |
|---|---|---|---|---|---|
| ntor=4, nzeta=12 | 4 | 95.0 ms | 0.450 ms | **211x** | 0.174 ms |
| ntor=4, nzeta=12 | 8 | 271.7 ms | 0.516 ms | **527x** | 0.197 ms |
| ntor=4, nzeta=12 | 12 | 543.1 ms | 0.631 ms | **860x** | 0.211 ms |
| ntor=mpol-1 | 4 | 64.2 ms | 0.414 ms | **155x** | 0.165 ms |
| ntor=mpol-1 | 8 | 643.6 ms | 0.681 ms | **945x** | 0.209 ms |
| ntor=mpol-1 | 12 | 2882.6 ms | 1.204 ms | **2393x** | 0.234 ms |

The gap to the symmetric twin closes from 388x-12312x to 2.5x-5.1x; what is
left is the `longdouble` dtype keeping numpy off BLAS, plus eight
contractions against the twin's six.

In situ, on the arguments the wout writer actually passes (captured from a
real solve): 376x-538x on `cth_like_free_bdy_lasym_small`, 8.6x-12.3x on
`up_down_asymmetric_tokamak` (small, and ntor = 0 collapses the n loop).

**Bit-parity.** Against a frozen copy of the loop, on the captured
wout-writer arguments of both decks (3 calls each, all agreeing):

| deck | field | scale | max abs diff | rel to scale |
|---|---|---|---|---|
| `cth_like_free_bdy_lasym_small` (nfp=5, ntor=4, ns=15, 16x20) | bsubu | 2.688e-02 | 2.78e-17 | 1.03e-15 |
| | bsubv | 5.869e-01 | 8.88e-16 | 1.51e-15 |
| `up_down_asymmetric_tokamak` (ntor=0, ns=17, 16x1) | bsubu | 3.863e-01 | 8.33e-17 | 2.16e-16 |
| | bsubv | 3.207e+01 | 1.42e-14 | 4.43e-16 |

1-7 ulp of the field scale, i.e. summation-order round-off. Random-input
sweeps across mpol/ntor/nzeta edge shapes (including mpol = 1 and ntor = 0)
stay in 1.9e-16 - 9.0e-16 scale-relative. `tests/test_wout_golden.py` passes
unchanged, including its `up_down_asymmetric_tokamak` case, which is the gate
that matters.

**Precision decision: extended precision kept, not dropped.** The
accumulation stays `np.longdouble` end to end, dtype for dtype with the loop.
It was worth checking what that costs, because `longdouble` keeps numpy off
BLAS: a float64 clone of the identical contractions runs **1.07x** faster on
the cth deck and **1.01x** on up_down, and lands the same 1-2 ulp from the
loop. There was nothing to buy, so nothing was traded.

Worth recording: on this host (macOS arm64) `np.longdouble` has float64
machine parameters — itemsize 8, eps 2.22e-16 — so the extended accumulation
only actually engages where long double is wider, e.g. x86-64 at 80 bits.
Keeping the dtype leaves that platform behavior unchanged too.

## Local verification

GitHub Actions is out of minutes; everything below ran on this machine with
`python3` (MacPorts 3.11), `pytest -n 2`.

- `tests/test_bootstrap.py` (both parametrized rows), `tests/test_wout_golden.py`,
  `tests/test_stability.py`, `tests/test_test_manifest.py` — **62 passed,
  10 skipped** in 7:23
- `tests/test_lasym_free_case.py`, `tests/test_lasym_free_convergence.py`,
  `tests/test_boozer_tables.py`, `tests/test_postprocess_currents.py`,
  `tests/test_geometry_fields.py` — **40 passed, 1 skipped** in 1:46
- `ruff check vmex tests examples benchmarks tools` — clean
- `mypy vmex` — clean (68 source files)

No `tests/test_nyquist*.py` exists; the lasym filter's gate is the golden wout
parity above.

### DIFF
--- a/tests/test_bootstrap.py
+++ b/tests/test_bootstrap.py
@@ -5,7 +5,8 @@ Zeff > 1; V2/V3 Zenodo QA/QH cross-check (<= 1% vs simsopt Redl curves,
 <= 10% RMS vs SFINCS interior, trapped-fraction parity; skips without the
 local dataset); differentiability of the double-where guards; traceable vs
 wout lane agreement; V4 ``<J.B>`` identity vs wout ``jdotb`` (<= 2% Zenodo,
-<= 1% solovev); V5 ``f_boot <= 1e-3`` on the published optima; V6 Picard
+<= 1% solovev and the LASYM up-down-asymmetric tokamak); V5
+``f_boot <= 1e-3`` on the published optima; V6 Picard
 convergence in <= 10 iterations; and the ``current_dofs`` driver with
 FD-checked AC/CURTOR Jacobian columns.
 
@@ -466,6 +467,20 @@ def eq():
     return equilibrium
 
 
+@pytest.fixture(scope="module")
+def lasym_eq():
+    """Up-down-asymmetric tokamak (LASYM = T, ntor = 0).
+
+    The deck exhausts NITER before FTOL — as its VMEC2000 golden does — but
+    the wout tables are what the wout lanes read, so convergence is not the
+    fixture's contract; carrying ``bmns``/``gmns`` is.
+    """
+    equilibrium = opt.solve_equilibrium(
+        VmecInput.from_file(DATA_DIR / "input.up_down_asymmetric_tokamak"))
+    assert bool(equilibrium.wout.lasym)
+    return equilibrium
+
+
 def test_state_lane_matches_wout_lane(eq):
     """redl_geometry_from_state agrees with redl_geometry_from_wout at
     discretization level (solver internal grid vs 64x65 wout synthesis)."""
@@ -585,9 +600,15 @@ def _interp_full(wout, surfaces, values):
                      np.asarray(values, dtype=float))
 
 
-def test_vmec_j_dot_B_state_lane_matches_wout_jdotb(eq):
-    """Traceable identity <J.B> vs the wout-engine jdotb (jxbforce.f) on the
-    solovev equilibrium: <= 1% (observed ~5e-5); wout-table identity ditto."""
+@pytest.mark.parametrize("fixture", ["eq", "lasym_eq"])
+def test_vmec_j_dot_B_state_lane_matches_wout_jdotb(request, fixture):
+    """Traceable identity <J.B> vs the wout-engine jdotb (jxbforce.f): <= 1%
+    (observed ~5e-5 solovev, ~9e-4 up_down); wout-table identity ditto.
+
+    The LASYM row is the regression gate on the ``geom=None`` synthesis: with
+    the sine families dropped the last assertion missed by 2.8e-3.
+    """
+    eq = request.getfixturevalue(fixture)
     surfaces = np.linspace(0.2, 0.9, 8)
     jv = np.asarray(bs.vmec_j_dot_B(eq.state, eq.runtime, surfaces=surfaces))
     jw = _interp_full(eq.wout, surfaces, eq.wout.jdotb)
--- a/tests/test_bootstrap.py
+++ b/tests/test_bootstrap.py
@@ -621,6 +621,18 @@ def test_vmec_j_dot_B_state_lane_matches_wout_jdotb(request, fixture):
     geom = bs.redl_geometry_from_wout(eq.wout, surfaces)
     jv_geom = np.asarray(bs.vmec_j_dot_B_from_wout(eq.wout, surfaces, geom=geom))
     np.testing.assert_allclose(jv_geom, jv_wout, rtol=1e-3)
+    # The two lanes are the same identity, so they must agree exactly, not
+    # merely within the gate above -- and at the axis too.  Every m != 0
+    # harmonic vanishes where the surface degenerates to a point;
+    # redl_geometry_from_wout zeroes them there and the geom=None lane did
+    # not, which put the two 4.0e-4 apart at s = 0 while they matched
+    # bit-for-bit everywhere else.
+    with_axis = np.concatenate([[0.0], surfaces])
+    axis_geom = bs.redl_geometry_from_wout(eq.wout, with_axis)
+    np.testing.assert_array_equal(
+        np.asarray(bs.vmec_j_dot_B_from_wout(eq.wout, with_axis)),
+        np.asarray(bs.vmec_j_dot_B_from_wout(eq.wout, with_axis,
+                                             geom=axis_geom)))
 
 
 @needs_zenodo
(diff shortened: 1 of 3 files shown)


rogeriojorge:
Blocked by a CI hole, not by a test gap — see #153.

`Changed executable lines (>= 95%)` reports `vmex/core/bootstrap.py (0.0%): Missing lines 741-742,751-753,757-760`. That is the whole `geom=None` branch of `vmec_j_dot_B_from_wout`, and it **is** covered — by `test_vmec_j_dot_B_state_lane_matches_wout_jdotb[eq]/[lasym_eq]` in `tests/test_bootstrap.py`.

The problem is that `tests/test_bootstrap.py` is never executed by any CI job. Its primary lane in `tests/manifest.json` is `pr-parity-b`, and no workflow invokes `pr-parity-b` — or eleven of the other thirteen primary lanes. 36 modules and 256 tests are outside pull-request CI. #153 wires them and adds the guard test that fails on `main`.

Nothing to change here. Merge #153 first and this gate should pass on a re-run.

rogeriojorge:
Rebased onto main now that #149 and #150 are in. Re-verified locally after the rebase: `test_bootstrap.py` + `test_wout_golden.py` + `test_postprocess_currents.py` — **37 passed, 8 skipped**; ruff and mypy clean.

Still waiting on #153 for the coverage gate — nothing to change here.

rogeriojorge:
Reviewed this rather than just rebasing it, and found one more thing — pushed here.

The two `<J.B>` lanes are the same identity, and after the sine-family fix they agree **bit-for-bit** from s = 0.2 to 0.9. At **s = 0 they were 4.0e-4 apart.**

Every `m != 0` harmonic vanishes at the magnetic axis, where the surface degenerates to a point. `redl_geometry_from_wout` zeroes them there; the `geom=None` synthesis did not, so it carried harmonics that should not exist — the `geom=` lane was the correct one, and the `rtol=1e-3` gate between them had been passing on it.

Both now agree exactly at the axis too, and the test asserts **array equality** rather than a tolerance, because two evaluations of one identity have no reason to differ at all.

Also rebased onto main, so the coverage gate now sees `tests/test_bootstrap.py` running — that module's primary lane (`pr-parity-b`) was one of the twelve #153 wired up, which is what was failing this PR before.

`tests/test_bootstrap.py` 24 passed, 8 skipped; ruff and mypy clean.

rogeriojorge:
The coverage gate's last two lines are closed, and chasing them was worth more than the number.

`nyquist.py:326,328` are the self-conjugate `dnorm1` half-weights. **They look like dead code and are not.** `mmax_force` is `mpol - 1` and `mnyq` is `ntheta2 - 1`, so the half-weight only engages when the grid under-resolves the retained modes — and `VmecInput` does not forbid that: `MPOL = 8` with `NTHETA = 10` is accepted. I had already written the justification for deleting them before I tested the coarse-grid case and found the branch live. That would have been a real mistake, so the test now pins the regime that reaches them.

**The docstring's scale contract was also overclaiming.** It said flatly that one pass is an idempotent projection. Measured: idempotent to **3.8e-16** whenever the force band lies strictly inside the grid Nyquist — every deck that resolves its own `MPOL` — but with the half-weights engaged a second pass moves the field by **1.7 relative**, not round-off. The docstring now names the regime it means, and the tests sit on both sides of that boundary.

Worth noting `filter_bsubuv_lasym` **had no direct test at all** — its only coverage was incidental, through the wout writer — so the property its own docstring claimed had never been gated. It is now.

`tests/test_bootstrap.py` 26 passed, 8 skipped; ruff and mypy clean.

Still gated on #162 (parity lane a is being OOM-killed at `-n 4`, which is my regression from #153, not anything in this PR).

PR #152, opened 2026-08-24
Ballooning: asymmetric states, a COBRAVMEC oracle, and an example

rogeriojorge:
Three things the ballooning lane was missing: asymmetric states, an external oracle, and an example. Also one wrong citation.

## LASYM

The eigenproblem does not change form for an asymmetric equilibrium. COBRAVMEC's `coeffs.f`, `getmatrix.f`, `geteigm.f` and `variat_eig_full.f` contain no `lasym` branches at all — only its Fourier sums do — and Correa-Restrepo (1978) / Dewar & Glasser, *Phys. Fluids* **26**, 3038 (1983) derive the operator for general 3-D systems with no symmetry assumption. So the work is in two spectral closures, and everything downstream (√g, the dual basis, **B**, ∇|B|, ∇α) follows by AD.

`_ballooning_context` now carries `rmns`/`zmnc`/`lmnc` alongside `rmnc`/`zmns`/`lmns`, and the PEST inversion takes the cos-parity λ. The three lanes that share this geometry each had five byte-identical closures (`coeffs`, `lam_fn`, `pos_fn`, `b_vector`, `modb_fn`); `stability.py` and `turbulence.py` are now built on one `_surface_closures` factory, so parity enters in a single place. **Correction:** an earlier revision of this description said all three were converted — `gammac.py` still carries its own copy here and is converted in #155.

**An asymmetric state has no α → −α parity.** That is why COBRAVMEC disables its own half-domain shortcut for `lasym` (`get_ballooning_grate.f:180`), and the default field lines now span the full `[0, 2π)` there.

Measured:

| gate | result |
| --- | --- |
| zero sine spectra through the asymmetric branch | **bit-identical** to the symmetric lane |
| same deck solved with `LASYM = T` | within **2.8e-13** of the symmetric lane (residual sine content 2e-16 relative) |
| α parity, symmetric state | holds to **7.4e-14** |
| α parity, asymmetric state | violated at **2.9e-3** |
| AD vs central difference, asymmetric lane | `rel = 2e-4` gate |

Turbulence and Γ_c keep explicit guards that name themselves. The geometry now reaches them, but neither has an asymmetric oracle yet, and unblocked-but-unverified is the silent-wrong class.

## COBRAVMEC parity

The lane had no external oracle. It needed no build: the STELLOPT binary reads a vmex-written wout directly — every field `order_input.f` asks for is already in the writer — and the eigenvalue converts **analytically**.

COBRA's third column is a *signed* growth rate, `+sqrt(-eigf)` unstable and `-sqrt(eigf)` stable, so `sign(grate)·grate²` is its eigenvalue under `λ = γ²μ0ρR0²/B0²`. vmex normalizes to `a_N`, `B_N`, so the ratio is `(a_N B0/(R0 B_N))²` — constant per equilibrium, with ρ and COBRA's own `amin` cancelling. Nothing is fitted.

solovev at `pres_scale = 3e4`, `k_w = 4`, `α = 60°`, relative deviation:

| s | 0.20 | 0.35 | 0.50 | 0.65 | 0.80 |
| ns = 49 | -4.3e-2 | -2.7e-3 | +2.5e-4 | +7.5e-4 | +6.3e-4 |
| ns = 99 | -6.9e-3 | -3.5e-4 | +4.1e-4 | +5.8e-4 | +4.7e-4 |

The innermost surface is COBRA's near-axis radial differencing, not a formulation difference — it falls sixfold when `ns` doubles while the outer surfaces sit at 6e-4 throughout. On `li383_low_res` (nfp = 3, β = 4.3%) the signs agree on all five surfaces, including which one is unstable, with absolute deviations at a uniform 1e-5 floor on values spanning 2e-5 to 1.4e-3.

Predicted factor 9.4416e-02 against measured 9.4475e-02 — **6e-4**. That agreement is the real content of the test: the normalizations were derived, not tuned.

## Turbulence: the guard came off, and it earned it

The GK geometry lane was symmetric-only. It is not any more, and not by deleting a guard — simsopt's `vmec_fieldlines` carries the sine-parity families independently (`vmec_diagnostics.py:748-761` splines `rmns`/`zmnc`/`lmnc`), so comparing against it on an **asymmetric** wout is a real external check. The pre-existing `test_geometry_matches_stability_conventions` only ever compared this module to `stability.py` — vmex against vmex.

Measured on `up_down_asymmetric_tokamak` at ns = 17/33/65: `gradpar` 2.6e-3 flat, `bmag` 5.9e-3 → 2.9e-3, `gds2` ~4.0e-2, drifts 5.8e-2 → 3.9e-2. **The symmetric row behaves identically**, which is the whole point — the parity handling adds no error of its own.

That comparison also settled something the module had asserted but never checked. **vmex and simsopt differ by ~3% in the drifts, permanently, and neither is wrong.** vmex evaluates the exact spectral |B| of the geometry; `vmec_fieldlines` reads the wout `bmnc` Nyquist table, which is band-limited (`xm_nyq` reaches 7 against a geometry `xm` of 3; the top two m bands hold 0.34% of the table). The drifts take a radial derivative of |B|, amplifying it by `1/hs`. That this is the band limit and not an error is measured, not asserted: **simsopt's own `gbdrift` moves 6.9e-3 between ns = 101 and 201 and vmex's moves 5.9e-3, while the gap between them plateaus at 3.1e-2** — it neither converges away nor grows. The test encodes that split: `gradpar` and the pressure term gated at 3e-3, drifts at a documented ceiling.

The pressure-term gate is normalized against the drift it contributes to rather than against itself, because that term is 8.7e-2 of the drift on li383 but 6.7e-4 on the low-beta asymmetric deck, where a self-relative gate would be measuring noise.

Γ_c keeps its guard on this branch; its asymmetric evidence is the exact reflection identity, which lands in #155.

## The example, and two defaults it exposes

`examples/optimization/QA_optimization_ballooning.py` optimizes a QA boundary against the growth rate, seeded on a configuration that is **Mercier-stable and ballooning-unstable** — the case the objective exists for, since the interchange criteria see nothing to fix.

Writing it surfaced two defaults that bias toward *false stable*, the dangerous direction for a stability constraint:

- **`zeta0s=(0.0,)`.** λ peaks at a configuration-dependent ballooning parameter (Gaur 2023, footnote 2). On that seed the single point reports **3.27e-3** where a five-point scan reports **4.42e-3** — a 26% under-report. A ζ0-blind objective drives the wrong quantity to zero. DESC scans 15 values; Gaur scans 15-21. The example scans, and a test pins that it does.
- **Softmax bias.** `ballooning_growth_rate` sits above `max λ` by up to `T·log(n_lines)`, which at the default `T = 0.05` over 12 lines is **0.12** — larger than published λ_max values (Gaur 2023 Fig. 6: 0.008-0.02). It is conservative, but the docs suggested targeting -0.01, which really targets λ ≈ -0.13. Now documented as the bound it is.

I have not changed the defaults themselves. That is a measurement question for the plan's 34.2, not something to slip into this PR.

## Attribution

The module cited "Sanchez, Hirshman, Ware, Berry & Batchelor, *J. Comput. Phys.* **161**, 589 (2000)". **That paper does not exist.** COBRA is Sanchez, Hirshman, Whitson & Ware, JCP **161**, 576 (2000); the VMEC-coordinate variant on p. 589 is Sanchez, Hirshman & Wong, *CPC* **135**, 82 (2001); Berry/Batchelor belong to PPCF **42**, 641 (2000).

The primary differentiable reference is now Gaur et al., *PPCF* **67**, 125015 (2025), not the 2023 JPP paper: that paper's printed Eq. (14) gives `f = |∇α|²/(B/B_N)²`, which is dimensionally inconsistent — λ would not scale as γ². **vmex implements the corrected form, not the typo**, as do Gaur's own code, DESC and COBRA.

## Verification

- `test_stability.py` — **35 passed** including the six new LASYM gates and the COBRA oracle (24 s with `RUN_FULL=1`)
- `test_turbulence.py` — 9 passed; the 3 failures are the pre-existing GKX `eig(enable_eigvec_derivs=)` break against jax 0.9.2 that #149 fixes, unrelated to this branch
- `test_gammac.py`, `test_test_manifest.py` — pass
- `ruff`, `mypy` (68 files), `check_docs_prose.py`, `sphinx -W` — clean
- example CI-smoke path: cost 1.327e-2 → 1.070e-2, λ_max 8.32e-3 → 7.24e-3

A full-budget production run of the example is still going; I will post its numbers here when it lands.

### DIFF
--- a/docs/reference/objectives.rst
+++ b/docs/reference/objectives.rst
@@ -474,8 +474,9 @@ MHD stability
 
 :mod:`vmex.core.stability` provides the infinite-n ideal-ballooning
 objective (plan R26h.h1) — a JAX port of the COBRA eigenproblem in the Gaur
-*et al.* (arXiv:2302.07673) formulation, with field-line coefficients per
-simsopt's COBRA-validated ``vmec_fieldlines`` conventions and a batched
+*et al.* formulation (*Plasma Phys. Control. Fusion* **67**, 125015 (2025),
+arXiv:2410.04576), with field-line coefficients per simsopt's
+COBRA-validated ``vmec_fieldlines`` conventions and a batched
 symmetric-tridiagonal ``eigvalsh`` solve:
 
 - :func:`~vmex.core.stability.ballooning_lambda` — the most-unstable
@@ -493,10 +494,36 @@ symmetric-tridiagonal ``eigvalsh`` solve:
             (ballooning_growth_rate, -0.01, 5.0)]   # keep λ_max below zero
    result = opt.least_squares(terms, inp, max_mode=4, jac="implicit")
 
+``examples/optimization/QA_optimization_ballooning.py`` runs this end to end
+on a seed that is Mercier-stable and ballooning-unstable — the case the
+objective exists for, since the interchange criteria see nothing to fix.
+
+Two defaults are worth setting deliberately rather than inheriting:
+
+- **Scan** ``zeta0``.  ``λ`` peaks at a configuration-dependent ballooning
+  parameter (Gaur *et al.* 2023, footnote 2), and the single-point default
+  ``zeta0s=(0.0,)`` under-reports it — by 26 % on that example's seed
+  (3.27e-3 at ``zeta0 = 0`` against 4.42e-3 over a five-point scan).  A
+  ``zeta0``-blind objective drives the wrong quantity to zero.
+- **Read the softmax as the bound it is.**  ``ballooning_growth_rate`` sits
+  above ``max λ`` by up to ``temperature × log(n_lines)``, which at the
+  default ``temperature=0.05`` over 12 lines is 0.12 — larger than published
+  ``λ_max`` values.  Target the bound (below zero is then sufficient for
+  stability) and report
+  :func:`~vmex.core.stability.ballooning_lambda`'s hard maximum, or lower the
+  temperature and accept a sharper gradient.
+
+Asymmetric (``lasym``) states are supported: the sine-parity ``R``/``Z``/``λ``
+spectra carry through the PEST angle map and the field-line geometry.  An
+asymmetric equilibrium has no ``α -> -α`` parity, so the default field lines
+span the full ``[0, 2π)`` there and ``zeta0`` should be scanned over both
+signs.
+
 Everything inside is JAX AD (geometry derivatives included), so it composes
-with both gradient modes; ``d(growth)/d(pres_scale)`` matches finite
-differences to 4.7e-9 in CI, and the objective destabilizes monotonically
-with pressure on the solovev family, in sign agreement with Mercier.  For
+with both gradient modes; ``d(growth)/d(pres_scale)`` matches a central
+difference at ``h = 1e-4`` to 4.7e-9 relative, and the objective destabilizes
+monotonically with pressure on the solovev family, in sign agreement with
+Mercier.  For
 interchange stability, combine with
 :func:`~vmex.core.optimize.magnetic_well` (traceable) or
 :func:`~vmex.core.optimize.mercier_stability_residual` (traceable); retain
--- a/examples/optimization/QA_optimization_ballooning.py
+++ b/examples/optimization/QA_optimization_ballooning.py
@@ -97,7 +97,12 @@ for max_mode, max_nfev in zip(MAX_MODES, MAX_NFEV):
     report(f"mode {max_mode}", equilibrium)
     print(f"max lambda = {worst_lambda(equilibrium):+.4e}")
 
-# The physics certificate is the resolved solve, not the optimizer's grid.
+# The physics certificate is the resolved solve, not the optimizer's grid, and
+# the difference is not small: a full run of this example reaches
+# max lambda = 2.2e-4 on the NS = 25 stage grid and 9.1e-4 when the same
+# boundary is re-solved at NS = 45.  The optimizer's number is optimistic by
+# a factor of four -- ballooning is radially stiff, so quote the resolved one
+# and add a stage at the certificate resolution if you need the margin.
 final_input = replace(inp, ns_array=np.array([FINAL_NS]),
                       ftol_array=np.array([1.0e-11 if ci_smoke else 1.0e-13]),
                       niter_array=np.array([8000]))
--- a/examples/optimization/QA_optimization_ballooning.py
+++ b/examples/optimization/QA_optimization_ballooning.py
@@ -1,0 +1,122 @@
+#!/usr/bin/env python
+"""Quasi-axisymmetric optimization against the infinite-n ballooning growth rate.
+
+The seed (``input.nfp2_QA_finite_beta``, beta = 2.7 %) is Mercier-STABLE and
+ballooning-UNSTABLE, which is the case a ballooning objective is for: the
+interchange criteria see nothing to fix.  The objective is the smooth upper
+bound ``ballooning_growth_rate``, so driving it below zero is a sufficient
+condition for every sampled field line to be stable.
+"""
+
+from dataclasses import replace
+from functools import partial
+import os
+from pathlib import Path
+
+import numpy as np
+from scipy.optimize import least_squares
+
+import vmex as vj
+from vmex import optimize as opt
+from vmex.core.stability import ballooning_growth_rate, ballooning_lambda
+
+SURFACES = np.linspace(0.1, 1.0, 6)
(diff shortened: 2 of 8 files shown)


rogeriojorge:
## Production run

Full budget (`MAX_MODES = [1, 2]`, stage `ns = 25`, certificate `ns = 45`), and the result is a mixed one worth stating plainly.

| | seed | mode 1 | mode 2 (ns 25) | certificate (ns 45) |
| max λ | +4.4210e-03 | +2.7029e-03 | **+2.2385e-04** | **+9.1242e-04** |
| QS total | 7.329e-05 | 1.851e-04 | 3.173e-04 | 3.149e-04 |
| min DMerc | +5.371e-06 | | | +1.175e-05 |
| β | 2.698% | | | 2.768% |

Two things to read out of that:

**λ_max falls 4.8× and does not reach zero.** The example is honest about it — it prints `still unstable`. Driving a 2.7%-β QA configuration ballooning-stable is not a two-stage job, and an example that claimed otherwise would be the interesting-looking kind of wrong.

**The optimizer's grid is optimistic by 4×.** Mode 2 reports 2.24e-4 at ns = 25; re-solving the *same boundary* at ns = 45 gives 9.12e-4. Ballooning is radially stiff, so the stage number is not the physics number. I have put that in the example as a comment with both figures, because anyone reading only the stage output would quote a margin they do not have.

The cost of stability was QS: 7.33e-5 → 3.15e-4, a factor of 4.3. Mercier improved on its own (5.37e-6 → 1.18e-5) and β was preserved. Eight figures written, `plot_wout` clean.

rogeriojorge:
Same blocker as #151, and this PR is where it shows worst — see #153.

`Changed executable lines (>= 95%)` reports `vmex/core/stability.py (26.8%): Missing lines 683-685,723,750-751,756-760,765-766,789-790,812-814,821-825,830-835,848-851,853,865,930-931,939-940,1032,1034` — essentially the whole `_surface_closures` / `_make_point_fn` / context block, plus the new tests' own lines.

Those lines are covered. The six LASYM gates and the COBRAVMEC oracle exercise all of them, and they pass locally. But `tests/test_stability.py`'s primary lane is `pr-parity-c2`, which no CI job invokes — so in CI the module runs only through three individually-named node ids in `pr-fast` and `pr-physics-field`, none of which touch this code.

Merge #153 first; this gate should then reflect the tests that actually exist.

rogeriojorge:
Production run of the example finished — the numbers I said I would post.

Seed `input.nfp2_QA_finite_beta` at beta = 2.698%, two stages (max_mode 1 then 2, 32 evaluations), optimizer grid NS = 25, certificate NS = 45:

| | seed | mode 1 | mode 2 | final (NS = 45) |
| max lambda | +4.4210e-03 | +2.7029e-03 | +2.2385e-04 | **+9.1242e-04** |
| QS total | 7.33e-05 | 1.851e-04 | 3.173e-04 | 3.149e-04 |
| min DMerc | +5.371e-06 | | | **+1.175e-05** |

**lambda_max falls 4.8x** and Mercier stays stable throughout — it more than doubles, so the ballooning term did not buy stability against interchange. QS degrades about 4x, which is the expected trade at `BALLOONING_WEIGHT = 200`; that constant is the knob.

One result worth reading off it, and the reason the example ends on a resolved solve rather than quoting the optimizer's own number: **the optimizer's NS = 25 grid reports lambda_max = 2.24e-4 where the resolved NS = 45 solve says 9.12e-4** — a 4x under-report in the direction that flatters the optimization. The configuration is still ballooning-unstable at the end. The example prints both, and does not claim otherwise.

Note this PR is blocked on #153 for the coverage gate, not on anything here.

rogeriojorge:
Correcting something I overstated when I opened this.

I wrote that the three field-line lanes "are now one `_surface_closures` factory". They were not: `stability.py` used it and `turbulence.py` and `gammac.py` still carried their own byte-identical `pos_fn`, `lam_fn`, `b_vector` and `modb_fn`. I checked because it is the kind of claim that is easy to write and easy not to verify.

Now made true for turbulence — pushed here, net a deletion, and the lane keeps its symmetric-only guard since the geometry reaching it is not the same as it being verified. `gammac.py` is converted in #155. The PR description is edited to match.

`tests/test_turbulence.py` + `tests/test_stability.py`: 36 passed, 3 skipped. The 3 failures are the pre-existing GKX `eig`/jax-0.9.2 break, identical before and after this change, and fixed in #153.

PR #153, opened 2026-08-24
Run the manifest primary lanes that no CI job was invoking

rogeriojorge:
`tests/manifest.json` makes every module declare exactly one primary PR lane, and `tools/test_manifest.py validate` enforces it — so the union of those lanes is the whole pull-request suite. The other half of that contract was never checked: **the workflows have to invoke all of them, and twelve of fourteen were invoked by nothing.**

```
pr-parity-a1  pr-parity-a2  pr-parity-b   pr-parity-c1
pr-parity-c2  pr-parity-c3  pr-parity-d   pr-examples
pr-external   pr-full-only  pr-gradient   pr-mirror-spline
```

That took **36 modules / 256 non-full tests out of pull-request CI entirely.** All 256 pass — they were simply never run (6m45s locally, `-n 4`). Among the 36 is `tests/test_test_manifest.py`, the module that validates the manifest, which is why nothing noticed.

Stale markers still in `test_workflow_selects_manifest_lanes` (`A1_FILES=`, `C2_FILES=`, `core-a-c)`) say what happened: the workflow was once organized around this partition, was refactored into nine `pr-physics-*` jobs driven by explicit node lists, and the record-level lanes were left behind as dead labels.

## Why this surfaced now

[#151](https://github.com/uwplasma/vmex/pull/151) fails `Changed executable lines (>= 95%)` at 67% — `vmex/core/bootstrap.py (0.0%): Missing lines 741-742,751-753,757-760`. That is the entire `geom=None` branch of `vmec_j_dot_B_from_wout`, which **is** covered by `test_vmec_j_dot_B_state_lane_matches_wout_jdotb[eq]/[lasym_eq]` — a test in `tests/test_bootstrap.py`, whose primary lane is `pr-parity-b`, which no job runs. The gate was reporting a CI hole as a test gap. [#152](https://github.com/uwplasma/vmex/pull/152) would have hit the same wall for `tests/test_stability.py` (`pr-parity-c2`).

## What this does

Four sharded `parity` jobs cover the twelve lanes:

| job | selectors | non-full tests |
| --- | --- | --- |
| `a` | `pr-parity-a1 pr-parity-a2 pr-parity-b` | 194 |
| `c1` | `pr-parity-c1 pr-parity-c3 pr-parity-d` | 540 |
| `c2` | `pr-parity-c2` | 351 |
| `misc` | `pr-examples pr-external pr-full-only pr-gradient pr-mirror-spline` | 104 |

Lanes overlap, so each job dedups its file list (`sort -u`) before running — a module selected by two lanes runs once. Coverage from the new jobs is combined into the changed-line gate, and `pr-gate` waits on them.

**On the grouping and the timeout.** I am not going to pretend I know the runner durations. My laptop currently has three physics jobs competing on it, so its numbers would be fiction. `timeout-minutes` starts at 45 and the grouping is a first cut by test count; the honest next step is to read the four durations off the first CI run here and rebalance from those. If a job is slow enough to bother you, splitting `c1` is the obvious first move.

## The guard

`test_every_primary_pr_lane_is_invoked_by_a_workflow` closes the loop. It fails on `main` naming all twelve lanes, and it reads both spellings a workflow can use — a literal `select <lane>` and a matrix `selector:` value, which may now list several. Without it this recurs the next time the workflow is reshaped.

Local: `ruff` clean, `tests/test_test_manifest.py` 6 passed (was 5; the new one fails on `main`).

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -160,6 +160,65 @@ jobs:
           path: coverage.physics-${{ matrix.lane }}
           retention-days: 3
 
+  parity:
+    name: Manifest parity lane (${{ matrix.lane }})
+    runs-on: ubuntu-latest
+    # The primary-lane partition is the whole pull-request suite; these four
+    # jobs are the part no other job was running.  Durations come from the
+    # runner, not a guess -- rebalance the grouping from the reported times.
+    timeout-minutes: 45
+    strategy:
+      fail-fast: false
+      matrix:
+        include:
+          - lane: a
+            selector: pr-parity-a1 pr-parity-a2 pr-parity-b
+          - lane: c1
+            selector: pr-parity-c1 pr-parity-c3 pr-parity-d
+          - lane: c2
+            selector: pr-parity-c2
+          - lane: misc
+            selector: pr-examples pr-external pr-full-only pr-gradient pr-mirror-spline
+    steps:
+      - uses: actions/checkout@v7.0.1
+        with:
+          fetch-depth: 0
+      - uses: actions/setup-python@v7.0.0
+        with:
+          python-version: "3.12"
+          cache: pip
+      - name: Install
+        run: |
+          python -m pip install -U pip
+          python -m pip install ".[coils,turbulence,freeb]" pytest pytest-cov pytest-xdist
+      - name: Add XLA compile headroom
+        run: |
+          sudo fallocate -l 12G /mnt/vmex-swap
+          sudo chmod 600 /mnt/vmex-swap
+          sudo mkswap /mnt/vmex-swap
+          sudo swapon /mnt/vmex-swap
+      - name: Fetch the golden digests
+        run: python tools/fetch_assets.py --bundle golden-v1
+      - name: Run the manifest lanes
+        env:
+          JAX_ENABLE_X64: "1"
+        run: |
+          FILES=""
+          for LANE in ${{ matrix.selector }}; do
+            FILES="$FILES $(python tools/test_manifest.py select "$LANE")"
+          done
+          # Lanes overlap, and a module selected twice would run twice.
+          FILES="$(printf '%s\n' $FILES | sort -u | tr '\n' ' ')"
+          pytest -q -n 2 -m "not full and not weekly" \
+            $FILES --durations=20 --cov=vmex --cov-report=term
+      - name: Stash coverage
+        run: mv .coverage coverage.parity-${{ matrix.lane }}
+      - uses: actions/upload-artifact@v7.0.1
+        with:
+          name: coverage-parity-${{ matrix.lane }}
+          path: coverage.parity-${{ matrix.lane }}
+          retention-days: 3
+
   device:
     name: Two-device placement and AD
     runs-on: ubuntu-latest
@@ -195,7 +254,7 @@ jobs:
     name: Changed executable lines (>= 95%)
     runs-on: ubuntu-latest
     timeout-minutes: 5
-    needs: [fast, physics, device]
+    needs: [fast, physics, parity, device]
     steps:
       - uses: actions/checkout@v7.0.1
         with:
@@ -211,7 +270,7 @@ jobs:
           merge-multiple: true
       - name: Combine coverage and enforce the mirror floor
         run: |
-          coverage combine coverage.device coverage.fast coverage.physics-*
+          coverage combine coverage.device coverage.fast coverage.physics-* coverage.parity-*
           coverage report --include='*/vmex/mirror/*' --fail-under=90
           coverage xml
       - name: Enforce changed-line coverage
@@ -223,7 +282,7 @@ jobs:
     if: always()
     runs-on: ubuntu-latest
     timeout-minutes: 2
-    needs: [quality, fast, physics, device, changed-coverage]
+    needs: [quality, fast, physics, parity, device, changed-coverage]
     steps:
       - name: Require every review gate
         env:
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -164,8 +164,15 @@ jobs:
     name: Manifest parity lane (${{ matrix.lane }})
     runs-on: ubuntu-latest
     # The primary-lane partition is the whole pull-request suite; these four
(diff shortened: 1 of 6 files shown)


rogeriojorge:
Two things the wiring immediately caught, which is the argument for it.

**`test_multigrid_ladder`'s two compile-count tests were failing wherever they ran** — `assert c1 > 0` with `c1 = 0`. The prelude installs its logging handler on the `jax` logger at WARNING and *then* imports vmex, and `vmex/__init__.py`'s `_configure_jax_logging` sets `jax_logging_level = "ERROR"` by default. That raises the logger from 30 to 40 and filters exactly the WARNING-level `"Compiling ..."` records the counter is built on. Measured: level 30 before the import, 40 after, zero records counted. Fixed with `VMEX_JAX_LOGGING_LEVEL=WARNING` in the measurement environment — vmex's own documented override — plus `VMEX_COMPILATION_CACHE=disabled` so "cold" means cold. Both pass now and measure what their docstrings claim. Pushed here.

**`test_turbulence.py::test_growth_rate_is_itg_critical_gradient_monotone` fails on the runner** with `gamma_lo = 0.0215` against `< 1e-6`. That one is not a CI artifact — it is a units defect in the GKX bridge, and it needs its own PR rather than a patch here. `turbulence.py:432` sets GKX's deprecated `R_over_LTi`, whose own deprecation message says the field is consumed as `a/L`, never `R/L`. The arithmetic confirms it: `_default_gradient_linear_params`'s defaults are `tprim = 2.49`, `fprim = 0.8`, which are the Cyclone base case `R/L_T = 6.9`, `R/L_n = 2.2` divided by `R/a = 2.77`. So `r_over_lt = 6.9` sets `tprim = 6.9` where 2.49 is intended — **2.77x too strong** — and the test's nominally-subcritical `r_over_lt = 1.0` becomes `tprim = 1.0` instead of 0.36, which is above marginal. Recorded as Phase 37; it needs jax >= 0.10.1 to verify locally, so #149 lands first.

I would rather land this PR with that one failure named and tracked than quietly exclude the module, since the whole point here is that not running a test is not the same as it passing.

rogeriojorge:
Pushed the turbulence units fix here rather than as a follow-up, because this PR is what makes that lane run in CI at all and a lane that arrives failing has not been turned on. Flagging it since it is a physics change beyond the wiring you approved — say the word and I will split it out.

`turbulence.py` set GKX's deprecated `R_over_LTi`/`R_over_Ln` from vmex's `r_over_lt`/`r_over_ln`. GKX's deprecation message on that field says the operator consumes `a/L` — its `tprim`/`fprim` — never `R/L`. The arithmetic settles it: `_default_gradient_linear_params` ships `tprim = 2.49`, `fprim = 0.8`, which are the Cyclone base case `R/L_T = 6.9`, `R/L_n = 2.2` divided by that case's `R/a = 2.77`. So `r_over_lt = 6.9` reached the operator as `tprim = 6.9` — **R/a times too strongly driven, on every turbulence evaluation this repo has made.**

Now divided by the equilibrium's own aspect ratio. On the shaped tokamak deck (aspect 2.643) the Cyclone drive lands at `tprim = 2.611`, next to GKX's default, and the growth-rate test's nominally subcritical `r_over_lt = 1.0` becomes 0.378 rather than 1.0 — which is why it was reporting `gamma_lo = 0.0215` against `< 1e-6`. `params_linear` stays the escape hatch for native `a/L`.

The new `test_drive_gradients_reach_gkx_as_a_over_l` pins the conversion with no solver call, since the defect was entirely in how the parameters were built, and it passes here. The growth-rate test that consumes it is still skipped on this host — jax 0.9.2 against the >= 0.10.1 floor #149 declares — so **that assertion is verified on the runner, not locally.** If it does not come back green I will say so rather than paper over it.

Also rebased onto main now that #149 and #150 are in.

rogeriojorge:
One more thing, from the runner's own numbers rather than a guess.

The four parity lanes measured **30m00s, 27m40s, 26m54s, 21m22s** at `-n 2`. They run in parallel with the nine physics jobs, whose slowest was 14m04s — so this PR roughly doubled total CI wall clock to 30 minutes, which is the ceiling you set.

The tests are not the problem; the per-job width is. Widened to `-n 4` on the same 4-vCPU runner, which the `fast` lane already uses. Recorded in the workflow comment so the fallback is not guesswork: if a lane runs out of memory instead of getting faster, go back to `-n 2` and split `pr-parity-c1` and `pr-parity-c2` in the manifest — `test_optimize` is ~20 min of the c1 group, `test_gammac` plus `test_scaling` ~12 min of c2.

Also worth recording: on the previous run, **lanes a, c1 and misc all came back green** once the multigrid compile-counter fix landed. Only c2 was red, and only for the turbulence units defect now fixed here. So 256 previously-unrun tests produced exactly two genuine defects, both now closed.

rogeriojorge:
The units fix is verified **here**, not just on the runner — I was wrong that it could only be checked in CI.

`_require_gkx` was skipping every gkx test on jax < 0.10.1, but only the eigenvector-weighted lanes need `enable_eigvec_derivs`. `turbulent_growth_rate` reduces the operator with `jnp.linalg.eigvals` and works on any supported jax. I measured which tests actually need the floor rather than assuming: with the guard relaxed, **exactly two** fail on the missing kwarg. Those two now carry `_require_gkx_eigenvectors`; the module goes from 13 skipped to **11 passed, 2 skipped** on this host.

That is also how the units defect survived — nine turbulence tests were dark on every machine below the floor, and the only thing that would have run them was CI, which was not running the lane at all. Two independent holes lining up.

With the fix, locally:

| | before | after | assertion |
| --- | --- | --- | --- |
| `gamma_lo` (`r_over_lt = 1.0`) | 0.0215 (CI) | **8.9e-15** | `< 1e-6` |
| `gamma_hi` (`r_over_lt = 6.9`) | — | **0.184** | `0.05 < g < 5.0` |

The traced path is exercised too: under `jit`, `tprim = 2.610926`, `fprim = 0.832469`, with a finite gradient through the new shape dependence (the drive is specified as R/L, so a/L moves with the aspect ratio — that is the correct semantics, and it is differentiable).

PR #156, opened 2026-08-24
Honor multigrid solve controls in implicit least squares

krystophny:
## Summary

The implicit least-squares lane did not materialize multigrid ladder controls supplied through `solve_kwargs` into its static implicit solver configuration. The finite-difference lane forwarded the same controls to `solve_equilibrium`, so a caller could request matched FD/implicit equilibria and silently get different implicit settings.

This patch extracts `ns_array`, `ftol_array`, and `niter_array` from `solve_kwargs` and applies them to the immutable input before constructing the implicit configuration. Other solver keywords remain forwarded to the host solve. `forward_ftol` / `forward_max_iterations` remain available for the newer scalar control path.

## Minimal reproducer and regression test

The regression is exercised directly through the public `opt.least_squares(..., jac="implicit", solve_kwargs=...)` API:

```bash
python -m pytest -q tests/test_implicit_solve_kwargs.py
```

The test requests a strict `ns=[7]`, `ftol=[1e-11]`, `niter=[1200]` ladder and then a deliberately different `ns=[5]`, `ftol=[1e-6]`, `niter=[5]` ladder. It asserts both the requested residual-space shape and that the computed residual norms differ. Before this patch both calls silently used the deck's `[5]`, `1e-10`, `1000` ladder and produced the same implicit residual. The new module is recorded in `tests/manifest.json` under the `pr-parity-d` lane.

The historical VMEX 0.1.0 benchmark that motivated this PR had a 1.5e-14 residual-value match but a 2.47% residual-Jacobian discrepancy against independent full-resolve FD after the residual-space mismatch was repaired. This PR fixes the solver-control propagation issue; the full QH Jacobian accuracy remains a separate follow-up check and is not claimed here.

## Verification

The targeted test passes with JAX x64 enabled. On the current checkout it completes in 126.97 s. `ruff check` is clean for the changed Python files. The repository-wide manifest check still reports the pre-existing `tests/test_tracing.py` ownership/selector inconsistency; that is unrelated to this PR.


krystophny:
Rechecked against fresh upstream `main` (`4ce66cb6575530765a40124419b2f0a2a5b8c02f`) on aCluster with JAX x64 and the strict ladder `ns=[16,50]`, `ftol=[1e-16,1e-11]`, `niter=[600,2000]`.

- Current `main`: FD residual/Jacobian shape `(3921, 8)`, implicit `(2721, 8)`.
- This PR: both lanes `(3921, 8)`, confirming the requested controls are propagated.
- The independent FD lane took 63.94 s; implicit took 344.46 s.
- The PR implicit residual differs from the independent FD residual by relative `4.99e-4`; the Jacobian differs by `4.927945%` relative Frobenius norm.

Conclusion: the control-propagation fix is a real current-main bug and this PR is not outdated. The remaining response/anchor discrepancy is a separate issue; no exact-gradient claim is made here. The full recheck artifacts are retained in the companion benchmark repository under `evidence/vmex-current-main-recheck/`.

krystophny:
Follow-up low-resolution one-column probe (`ns=8`, `mpol=ntor=2`, one QH surface) keeps the 1,793-row spaces identical but still shows a 1.5677% relative error in the selected implicit-vs-full-resolve Jacobian column and residual-vector L2 difference `1.1456e-3` (39.92 s FD, 288.49 s implicit). This survives downscaling; it is likely an anchor/stopping-point or response-oracle issue, but we have not isolated a safe code fix. The control-propagation patch remains valid; exact-gradient claims remain rejected pending a separate MRE/fix.

rogeriojorge:
Reviewed against the repo's standards. **The fix itself is right and I'd take it.** Three things to resolve first.

## Verified locally

- Applies cleanly to current `main` (`89819835`); `ruff` and `mypy` (68 files) clean.
- `tests/test_implicit_solve_kwargs.py` passes with the patch (98 s) and **fails without it** — I reverted just the `optimize.py` hunk and it fails at line 66 on `got.fun.shape == expected.shape`. That makes it a genuine regression test, not a restatement.
- `vmex/resources/input.nfp4_QH_warm_start` exists, so the tool's default path is valid.
- The interaction I was most worried about is handled correctly: `_least_squares_implicit` ends with a diagnostic re-solve at `optimize.py:3410` using `**solve_kwargs`, and since the ladder keys were popped out and materialized into `inp`, that solve gets them via `result.input` — no double application, nothing dropped. The `dict(solve_kwargs or {})` copy also correctly avoids mutating the caller's dict.

## Three things

**1. CI has never run on this PR — zero checks.** "All tests passing" can't be claimed yet. Worth noting that `tests/test_implicit_solve_kwargs.py` is a new module with no `tests/manifest.json` record, and this repo enforces exactly-one-primary-lane ownership; it will need a record. #153 (in flight) additionally turns on twelve manifest lanes that no CI job was invoking, so the coverage gate here will behave differently once that lands.

**2. `tools/reproduce_qh_implicit_jacobian.py` should not ship.** 131 lines whose only consumer is the PR description. It is a reproducer, and the repo's standard is no scaffolds or testbeds, and new files only with a compelling rationale — the regression test already carries the fix. It also hardcodes `"vmex": "0.6.0-source"` in its provenance payload, which is a version string that will silently go stale, and emits a `schema: 1` JSON that nothing in the repo reads. The reproducer command belongs where it already is: the PR body.

**3. The test asserts the implementation where it could assert the behavior.** `assert np.asarray(got.input.ns_array).tolist() == [7]` checks that the input object was rewritten — which is *how* the fix works, not *what* it buys. A refactor achieving the same end differently would fail it for no reason. The behavioral version is cheap: run the implicit lane twice with `ns_array=[5]` and `[7]` and assert the residuals differ. Before the fix both were the deck's ladder and would be identical; after, they must not be. That is the property worth locking.

## Smaller notes

- The comment block at `optimize.py:2413-2423` is genuinely good — it says why `device` is deliberately excluded, which is the part a future reader would otherwise get wrong.
- Not covered: `solve_kwargs` ladder keys combined with `hot_restart=True`. The seed cache is keyed on the config, and the fallback at 3413 exists for "ns-mismatched seed", so it is probably fine — but it is the obvious edge case and it is untested.
- The PR body is careful about what it does not claim (the QH Jacobian accuracy follow-up). That is the right call and I'd keep it that way.

rogeriojorge:
Re-reviewed after your updates — all three points are addressed, and the CI I approved came back **19/19 green**, so this now has the test evidence it was missing when I first looked.

- `tests/manifest.json` record added, so the new module has an owning lane. Worth knowing that #153 landed in the meantime and turned on twelve primary lanes that no CI job had been invoking, so that record is now load-bearing rather than bookkeeping.
- `tools/reproduce_qh_implicit_jacobian.py` dropped.
- The behavioural assertion is in: `ns_array = [5]` against `[7]` must produce different residuals. That is the property worth locking — before the fix both lanes silently used the deck's ladder and would have been identical — and it survives a refactor that achieves the same end differently, which the `got.input.ns_array == [7]` check alone would not.

I'll merge this with `--rebase` rather than squash so your three commits and their authorship are preserved on `main`.

One thing outside your change, so you are not surprised by it: parity lane `a` is currently being OOM-killed at `-n 4` on some runs — my regression from #153, fixed in #162, which goes in first.

krystophny:
Added the hot-restart coverage Rogerio flagged and removed the remaining implementation-specific `got.input.*` assertions. The regression now checks finite, different residuals for `[7]` versus `[5]` ladders through the public implicit API. Targeted test passes (149.5 s) and pre-commit passes.

PR #159, opened 2026-08-24
Implement every remaining VMEC2000 profile parameterization

rogeriojorge:
Stacked on #158. Enumerated from `profile_functions.f`'s `SELECT CASE` labels rather than from memory — `pmass` has nine explicit cases plus `power_series` as `CASE DEFAULT`, `piota` six plus the default, `pcurr` sixteen plus the default. **vmex was missing seven.** Four land here, and **`PMASS_TYPE` and `PIOTA_TYPE` are now complete.**

| kind | lanes | |
| --- | --- | --- |
| `two_power_gs` | pressure, current | `two_power` × six Gaussian peaks in `c[3:21]` as (amplitude, centre, width) triples. Pressure evaluates it directly; current integrates it with the same 10-point Gauss-Legendre rule as `two_power` |
| `two_Lorentz` | pressure | two Lorentz terms, each shifted and normalized to 1 at the axis and 0 at the edge |
| `rational` | pressure, iota, current | `c[0:10]` over `c[10:21]`, both by Horner exactly as the Fortran loops run |
| `nice_quadratic` | iota | `c0` and `c1` are the axis and edge values; `c2` is the midpoint's departure from the straight line between them |

## Coverage after this

| | vmex | VMEC2000 |
| --- | --- | --- |
| `PMASS_TYPE` | 10 | 10 ✅ |
| `PIOTA_TYPE` | 7 | 7 ✅ |
| `PCURR_TYPE` | 14 | 17 |

## What is deliberately *not* implemented

The three `sum_cossq_*` current forms. They are analytic piecewise integrals of cos² windows with Heaviside boxcar windowing — Joachim Geiger's 2018 additions — and are substantially more work than the four above. They now raise `NotImplementedError` naming the supported set, **because a wrong answer is worse than a refusal.** `test_every_vmec2000_profile_type_is_selectable` pins both halves: everything claimed is accepted and finite, and those three still raise.

## Verified twice, as with #158

**Formula**, against literal transcriptions of `profile_functions.f` and `functions.f`: exact (0.0) for `two_power_gs`, `rational` and `nice_quadratic`; 7.4e-17 for `two_Lorentz`. `rational`'s zero-denominator sentinel is bit-equal to Fortran `HUGE`.

**End to end**, live `xvmec2000` on `li383_low_res` — `two_Lorentz` pressure `presf` **1.4e-16**, `iotaf` 2.9e-13, `betatotal` 6.8e-14; plus a `rational` iota row, alongside #158's two `sum_atan` rows. Each row runs on the lane where its type is actually consumed and asserts VMEC echoed the type it used, so a fallback to `power_series` cannot pass silently.

## Two details worth a look

- **The `HUGE` sentinel.** VMEC returns `HUGE(real64)` where the `rational` denominator vanishes — not NaN, not infinity. That value reaches a wout and a caller can see it, so it is parity, not an implementation detail. Reproduced and tested.
- **The guards were hardcoded tuples** in `pressure()` and `iota()` that had to be edited in step with the dispatch table — the kind of duplication that goes stale. They are now `frozenset`s declared next to it, and the error names what is supported.

`tests/test_profiles.py` 47 passed; four live parity rows pass under `--run-vmec2000`; `ruff`, `mypy`, `check_docs_prose`, `sphinx -W` clean. Docs updated in `input-file.rst` and the `profiles` module table.

### DIFF
--- a/docs/howto/profiles.md
+++ b/docs/howto/profiles.md
@@ -89,6 +89,10 @@ $c_0 + c_1 + c_5 + c_9 + c_{13} + c_{17}$, which is the limit only when the
 scales and the $(1-x)$ exponents are positive; vmex reproduces that as
 written.
 
+vmex implements every parameterization VMEC2000 offers: ten for
+`PMASS_TYPE`, seven for `PIOTA_TYPE`, seventeen for `PCURR_TYPE`. A name
+outside those raises rather than falling back to a default.
+
 Every profile key, default, and accepted `*_TYPE` string:
 {doc}`/reference/input-file` (Pressure profile / Current and iota sections).
 The evaluation code is {mod}`vmex.core.profiles`.
--- a/docs/reference/input-file.rst
+++ b/docs/reference/input-file.rst
@@ -167,8 +167,9 @@ Pressure profile
      - Meaning
    * - ``PMASS_TYPE``
      - ``power_series``
-     - one of ``power_series``, ``two_power``, ``gauss_trunc``,
-       ``cubic_spline``, ``akima_spline``, ``line_segment``, ``pedestal``,
+     - one of ``power_series``, ``two_power``, ``two_power_gs``,
+       ``two_Lorentz``, ``gauss_trunc``, ``rational``, ``cubic_spline``,
+       ``akima_spline``, ``line_segment``, ``pedestal``,
        ... (see :mod:`vmex.core.profiles`)
    * - ``AM``
      - zeros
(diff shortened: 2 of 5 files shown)


rogeriojorge:
Pushed the remaining three, so this PR now completes the set rather than stopping short of it.

**vmex implements every parameterization VMEC2000 offers: `PMASS_TYPE` 10/10, `PIOTA_TYPE` 7/7, `PCURR_TYPE` 17/17.**

The `sum_cossq_*` forms (J. Geiger, 2018) build a current density from windowed cos² waves and integrate it analytically, so `I(s)` is closed form — a window the argument has already passed contributes its whole area and at most one is partial. `sum_cossq_s` places `c[0]` windows evenly in `s`; `sum_cossq_sqrts` places them in `sqrt(s)`, which changes the closed form because the integral picks up the extra `x` weight; `sum_cossq_s_free` takes seven freely placed windows with individual widths from `c[3i:3i+3]`.

**Bit-exact** against literal transcriptions of `profile_functions.f` — **0.0** relative error on all five coefficient sets, on grids that cross several window boundaries, which is where the accumulate-completed-windows bookkeeping would otherwise hide an error.

Two decisions worth your eye:

- **The wave count `c[0]` is read as a Python `int`.** It sets how many terms the sum has, so it fixes the shape. It is a count and never an optimized coefficient, so this costs nothing real — and a tracer there now gets a message saying so, rather than a shape error from deeper in the trace. Out-of-range counts raise where VMEC `stop`s.
- **A Fortran quirk is reproduced, not smoothed.** The window test is a strict `>`, so at `x = 0` the first window already counts as *fully* integrated and `I(0)` sits **above** `I` of the next sample. I found this by adding an `I(0) = 0` assertion that failed. It is real behaviour, the transcription comparison pins it, and it is why the monotonicity check starts at the second sample. I'd rather match VMEC than have a prettier curve.

Also: `test_every_vmec2000_profile_type_is_selectable` now asserts the counts (10/7/17), so a case added upstream is noticed instead of silently missing.

`tests/test_profiles.py` **53 passed**; `ruff`, `mypy`, `check_docs_prose`, `sphinx -W` clean.

PR #148, opened 2026-08-24
Anchor the implicit gradient lane at a refined fixed point

rogeriojorge:
Supersedes the strict `xfail` added to
`test_lasym_3d_stability_gradients_vs_frozen_path_fd` in #147: that test now
passes, and at a gate 100x tighter than the one it was failing.

## Root cause

The implicit function theorem gives the derivative of the equilibrium **at a
root of the residual**, and the host solver does not stop at one. `ftol`
gates `fsqr + fsqz + fsql`, the sum of SQUARES of the force, so a solve the
host reports converged returns with `|F| ~ sqrt(ftol)` — on every deck in the
gradient lane:

| deck | `ftol` | `\|F(P(x*), p)\|` at the host stop |
| --- | --- | --- |
| `solovev` | 1e-14 | 3.696e-08 |
| `li383_low_res` | 1e-13 | 9.374e-08 |
| `up_down_asymmetric_tokamak` | 1e-12 | 1.295e-07 |
| `basic_non_stellsym_simsopt` | 1e-12 | 2.673e-07 |

Where `dF/dz` carries a small singular value that residual is a large
*displacement*: on `basic_non_stellsym_simsopt` the lasym m=1 direction has
`sigma = 1.473e-04`, so `|F| = 2.7e-07` sits `1.804e-03` from the root
(`|z*| = 1.65e+01`). Newton-polishing onto the root moves
`sum DMerc[2:-1]` by 9.4e-09 — 1.7x the whole `h = 1e-4` FD numerator.

Value, cotangent and linearization were all read off that point. The
frozen-path FD is the accurate side: its own Newton endpoints are roots, so
it measures `dm/dx` there as well.

## The fix

`_refined_state` Newton-refines the host state onto the root of the frozen
residual, in the host callback — the one place every lane reads the state
from — so all three move together. The knob is
`ImplicitConfig.refine_tol` (default 1e-10, `inf` disables), alongside the
`adjoint_tol` family and plumbed through `make_config` and `im.run`.

* The projector confines the correction to the evolved dofs, so the refined
  state solves exactly the equations the host solver iterates, only closer.
* A refinement that does not improve the residual — a stalled Krylov step, a
  non-finite iterate — leaves the host state untouched. The anchor is an
  accuracy gain, never a precondition for returning a gradient.
* The steps are inexact-Newton (forcing term 1e-6, three steps, 20 GCROT
  cycles each): the step only has to land `|F|` under `refine_tol`, and
[...]

PR #160, opened 2026-08-24
Release 0.7.0

rogeriojorge:
Version bump only. Cuts 0.7.0 on the merged content: #136-#150 and #158.

Headline is `sum_atan` (#158) and the refined-anchor implicit gradient fix (#148). Full notes go on the GitHub Release.

Verified: wheel and sdist build clean at 0.7.0; `vmex.__version__` reports 0.7.0, which is what the `publish-pypi` workflow validates against the tag.

### DIFF
--- a/.github/workflows/publish-pypi.yml
+++ b/.github/workflows/publish-pypi.yml
@@ -81,6 +81,10 @@ jobs:
       - name: Install and solve outside the source tree
         env:
           ARTIFACT: ${{ matrix.artifact }}
+          # The tag, minus its leading v.  The validate job above has already
+          # checked it against pyproject, so this pins the *installed*
+          # artifact to the same version without a second source of truth.
+          EXPECTED_VERSION: ${{ github.event.release.tag_name || github.ref_name }}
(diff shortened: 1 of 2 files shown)


rogeriojorge:
Pushed a second commit here, because cutting the tag without it would have produced a release that looked complete while PyPI got nothing.

`publish-pypi.yml`'s wheel and sdist smoke test asserted `vmex.__version__ == "0.6.0"` as a literal. On a v0.7.0 tag the `verify` job fails, and `publish` is gated on it — so the GitHub Release would have been created and the package would silently not have shipped.

It now reads the tag minus its leading `v`. The `validate` job above already checks that tag against `pyproject.toml`, so this pins the *installed artifact* to the same version without adding a second source of truth, and it cannot go stale at the next release the way the literal did.

Also rebased onto main now that #159 is in. Verified locally: wheel and sdist build clean at 0.7.0, `vmex.__version__` reports 0.7.0, and the workflow still parses (`build`, `verify`, `publish`, `mark-latest`).

PR #167, opened 2026-08-29
Add branch-preserving strong-force polish driver

rogeriojorge:
## Summary
- add an immutable fixed-boundary polish configuration, result, and machine-readable report
- drive the exact anchored low endpoint with SOLVAX adaptive PTC continuation and physical Jacobian/admissibility checks
- add secant-seeded, bordered pseudo-arclength tangents near folds and typed continuation/certification failures
- retain the measured legacy inverse as the default; include a bounded neighboring-mode block variant as an explicitly experimental option
- expose the advanced driver lazily without changing existing solve behavior

## Evidence
- Ruff and mypy pass for the new source and tests
- focused polish/preconditioner module: 29 passed
- package API module: 21 passed
- strict Sphinx HTML build passes
- bounded Solovev production diagnostic: deterministic equation orientation moved the accepted branch from alpha about 0.51 to about 0.995 and the SOLVAX dynamic-parameter compile fix reduced the attempt from 347 s to 91 s

## Claim boundary
This PR provides and validates the driver/control-flow layer. It does not claim that the coarse lifted Solovev stress case reaches the certified alpha=1 root: ns=13-51 lifts retain large near-axis second-derivative defects, and the experimental mode-band preconditioner did not reduce total Krylov work. Public solve polish integration, p/h refinement, and a certified production case remain follow-up gates. Existing unpolished solve behavior is untouched.

### DIFF
--- a/docs/explanation/high-order-force-balance.rst
+++ b/docs/explanation/high-order-force-balance.rst
@@ -184,3 +184,46 @@ high-order angular coupling.  It is a right preconditioner, not the polishing
 operator, and it never assembles a dense high-order Jacobian.  The subsequent
 p-level hierarchy may wrap it as the coarse solve without changing this
 contract.
+
+The strong endpoint is first normalized by its frozen initial RMS and then by
+a deterministic matrix-free estimate of the low-inverse/strong-Jacobian
+stiffness.  This positive scalar leaves the target root unchanged while making
+the continuation endpoints comparable.  The estimate uses a fixed probe and a
+bounded number of JVP/low-inverse applications; it is stored as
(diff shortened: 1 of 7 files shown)


rogeriojorge:
Closing as experimental evidence under the DKX 3 admission gate. The driver reaches only the documented near-end continuation state (about alpha=0.995), not an independently certified production alpha=1 root. Generic/control-flow tests do not override that missing scientific endpoint. The later native-coordinate and conditioning drafts preserve the useful diagnostics; a replacement may be promoted only after a cheaper full endpoint and independent certificate pass.

PR #168, opened 2026-08-29
Expose opt-in polishing on fixed-boundary solve

rogeriojorge:
## Summary
- add False, True, and auto polish modes to the single-grid fixed-boundary solve
- preserve the existing legacy state and wout-compatible arrays exactly
- attach the native high-order equilibrium, independent force report, and polish report in new default-None result fields
- run the same fixed-point refinement used by implicit differentiation before lifting
- add the radial-degree control and validate API inputs before work begins

## Evidence
- Ruff and mypy pass
- real five-surface auto-polish integration smoke passes and preserves legacy/native shapes
- invalid-mode unit test passes
- strict Sphinx HTML build passes
- the preceding driver module has 29 passing focused tests and package API has 21 passing tests

## Claim boundary
This PR exposes only the non-differentiable fixed-boundary solve path. It does not claim polished implicit gradients; that requires the separate polished-root tangent and adjoint registration. Existing calls default to polish disabled.

### DIFF
--- a/docs/reference/api/advanced.rst
+++ b/docs/reference/api/advanced.rst
@@ -33,6 +33,12 @@ High-order correction transfer and preconditioner
 .. automodule:: vmex.core.polish_driver
    :members:
 
+The single-grid :func:`vmex.solve` entry point accepts ``polish=False``
+(unchanged behavior), ``polish=True`` (required correction), or
+``polish="auto"`` (skip an already-certified state).  Polished data is attached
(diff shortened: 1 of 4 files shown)


rogeriojorge:
Closing with its uncertified parent stack. This public opt-in API depends on #167, whose production alpha=1 root is not independently certified. Exposing the path before the scientific endpoint passes would exceed the stated claim boundary. Preserve the implementation as experimental evidence and re-cut only on an admitted root foundation.

PR #169, opened 2026-08-29
Add implicit derivatives for polished strong roots

rogeriojorge:
Stacked on #168; the continuation API comes from SOLVAX #87.

This PR adds the optimization-facing derivative layer for an alpha=1 strong-force root. It provides explicit matrix-free tangents and adjoints plus a custom VJP that differentiates the converged equation rather than replaying continuation. The local strong-force chart and positive normalization are frozen, while geometry, lambda, toroidal/poloidal flux profiles, and pressure remain differentiable.

The low endpoint linearizes as D A, so the transpose preconditioner explicitly applies D^-1 A^-T using the same stored block factors. Nonconvergence has a typed StrongForceLinearSolveError in eager execution and NaN poisoning under tracing.

Validation:
- 34 polishing tests pass, including full tangent/adjoint dot-product identity and custom-VJP equality
- ruff and mypy pass
- package lazy-export checks pass
- strict Sphinx documentation passes
- exact quality/package/docs workflow passes locally

Clean Apple M4 structural benchmark with persistent compilation cache disabled:
- tangent: 4.04 s cold, 3.94 ms warm
- adjoint: 4.99 s cold, 5.78 ms warm
- custom VJP: 5.46 s cold, 5.72 ms warm
- tangent/adjoint duality relative error: 1.77e-11
- custom VJP relative squared error: 0

The benchmark and documentation call this a full-rank structural derivative gate, not a certified production-equilibrium scaling claim.

PR #170, opened 2026-08-29
Cut QA optimization startup time and memory

rogeriojorge:
## Summary

- construct the fixed-boundary implicit DOF mask exactly from mode-table invariants, while retaining the numerical VJP mask as an independent oracle and for coupled free-boundary maps
- batch raw block-Jacobian VJP probes according to the public chunk setting instead of retaining every output-row tape at once
- move the canonical QA example to the mathematically identical scalarized objective and one reverse implicit adjoint, avoiding the 6723 x 48 pointwise Jacobian and three separate decision-width compilations
- add a reproducible cold/warm startup benchmark and clean M4 artifacts for both scalar and least-squares lanes

## Measured result (M4, cache disabled, 48 DOFs)

- scalar lane: 30.62 s cold startup, 2022.95 MiB peak RSS
- one least-squares stage: 40.45 s cold startup, 2776.08 MiB peak RSS
- former three-stage canonical example baseline: about 117 s startup and 3.73 GiB peak RSS
- five scalar L-BFGS-B iterations reduced the exact objective from 21.4488 to 0.4825 (97.8 percent)

The scalar and least-squares seed values agree to 2.8e-8 absolute; both use the same weighted residual tuples.

## Validation

- exact CI quality, package, prose, and strict Sphinx pipeline
- QA example CI smoke end to end, including final equilibrium and artifacts
- fixed-boundary analytic mask equals the numerical VJP oracle for 2-D symmetric, 3-D symmetric, and 2-D asymmetric cases; 3-D asymmetric equality remains in the full test lane
- raw block factors agree across probe chunk sizes; forward, transpose, and finite-difference response tests pass
- reproducible artifacts: benchmarks/qa_optimization_startup_scalar_m4.json and benchmarks/qa_optimization_startup_least_squares_m4.json

Stacked on #169.

### DIFF
--- a/benchmarks/qa_optimization_startup.py
+++ b/benchmarks/qa_optimization_startup.py
@@ -1,0 +1,158 @@
+#!/usr/bin/env python
+"""Measure cold QA problem startup, warm gradient latency, and peak RSS."""
+
+from __future__ import annotations
+
+import argparse
+from dataclasses import replace
+from importlib import metadata
+import json
+import os
+import platform
+from pathlib import Path
+import resource
+import statistics
+import time
+
+# Cold means no executable restored from a previous process.
+os.environ["VMEX_COMPILATION_CACHE"] = "disabled"
+
+import jax.numpy as jnp
+import numpy as np
+
+import vmex
+from vmex import optimize as opt
+
+from _provenance import assert_repo_vmex, file_sha256, git_state
+
+
+REPO = Path(__file__).resolve().parents[1]
+DATA = REPO / "examples" / "data" / "input.minimal_seed_nfp2"
+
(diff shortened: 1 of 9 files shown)


rogeriojorge:
Superseded by clean standalone PR #177, merged to `main` at `bca15514d5d60f20e670d6807ae751ba85fdb8a1` after every API, package/docs, device/AD, representative-physics, manifest-parity, changed-line coverage, and aggregate PR gate passed. The admitted branch contains the QA startup/memory work without depending on the uncertified #167-#169 root stack. Closing this stacked PR; its earlier performance history remains available here.

PR #171, opened 2026-08-29
Dealias and scale the corrected strong root

rogeriojorge:
## Summary

- preserve the branch-connected strong-force continuation driver and its implicit derivative path on current `main`
- overintegrate the strong residual and project it back to the square spline coefficient space
- retain typed alpha-zero, continuation, and pseudo-arclength failure reports

## Restack

This draft is now based directly on `main`. It excludes the closed QA stack (superseded by #177) and all WSL changes. The later scientific slices remain #174, #175, #176, and #178.

## Current claim boundary

This is an experimental integration base, not a merge candidate. The production diagnostic still fails the alpha=1 rank/certificate gate, so the public polish and implicit APIs remain draft-only.

## Focused verification

- polish continuation compatibility mock: pass
- aggregate polish suite is run on the final stacked head
- full hosted matrices are intentionally deferred until a scientific merge candidate exists

PR #174, opened 2026-08-29
Regularize the strong-force radial lift

rogeriojorge:
## Summary

- use an overdetermined degree-3/5/7 spline fit for the legacy-to-continuum lift instead of interpolating first-order radial mesh noise
- certify stable reconstruction before allocating the temporary legacy-coordinate chart
- keep the square polisher on the explicit compatibility chart until the following native-spline slice

## Measured reconstruction

For the ns=11 Solovev diagnostic, the overdetermined default reduced the lifted absolute force from `4.97e7` to `97.3 N/m^3`, radial quadrature change from `0.807` to `3.78e-5`, and improved the minimum signed Jacobian from `0.258` to `5.41`.

## Restack and claim boundary

This draft is stacked directly on #171; it no longer contains WSL changes. It improves the initial representation but does not claim completion of the corrected strong root. The independent alpha=1 certificate remains mandatory.

### DIFF
--- a/benchmarks/polish_implicit.py
+++ b/benchmarks/polish_implicit.py
@@ -31,6 +31,7 @@ from vmex.core.polish_implicit import (
     strong_root_adjoint,
     strong_root_tangent,
 )
+from vmex.core.radial_basis import BSplineBasis
 from vmex.core.strong_force import lift_high_order_state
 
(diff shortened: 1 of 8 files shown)


rogeriojorge:

PR #175, opened 2026-08-29
Use native spline coordinates for strong roots

rogeriojorge:
## Summary

- replace the temporary equal-size legacy chart with independent native spline coordinates
- build bounded local (m, |n|) image bases for the constrained high/low transfer, including coupled 3-D m=1 modes
- propagate the native coordinates through primal, tangent, adjoint, custom-VJP, mode-block, and benchmark paths
- intersect spline support with the explicit evolved low mask so stale or zero masks cannot manufacture free coordinates

## Validation

- 35 polishing, preconditioner, tangent, adjoint, and custom-VJP tests pass
- 15 independent strong-force oracle tests pass
- Ruff and git diff checks pass
- clean structural benchmark: 23/23 endpoint rank, 2.59e5 condition number, 0.287 ms warm residual, 0.427 ms warm JVP

## Draft promotion gates

This is intentionally a draft stacked on #174. Production Solovev diagnostics expose two issues that must be fixed before promotion:

1. Mode scaling: mpol=6 has 117 coordinates but global rank 39 at relative 1e-8 because otherwise full-rank Fourier blocks differ by many physical scales. Invertible Ruiz equilibration recovers 110 directions at 1e-8 and 116 at 1e-10; one block-local basis direction still needs conditioning.
2. Preconditioning and branch completion: the inherited legacy inverse is not exact in the smaller smoothing image. Ordinary mode-block continuation reaches alpha=0.9989308; dense diagnostic arclength approaches alpha=0.9999816 without crossing. The independent certificate is not weakened. Shared bounded mode-block primal/transpose factors and a certified endpoint formulation remain required.

The refreshed derivative artifact records the current regression honestly: tangent warm time improves to 3.51 ms, while adjoint/custom-VJP rise to 16.06/16.66 ms. Those transpose costs are a promotion gate, not a claimed win.


rogeriojorge:

PR #176, opened 2026-08-29
Reuse bounded mode blocks for polish adjoints

rogeriojorge:
## Summary

- move the bounded Fourier mode block pencil into the shared strong-root layer
- add exact-transpose block applications for implicit adjoints
- allow factors to be built at the converged correction and reused by tangents, adjoints, and the custom VJP
- make mode blocks the measured polishing default while retaining the inherited legacy inverse as an explicit fallback
- refresh the clean derivative artifact and documentation

## Measured result

On the clean 23-coordinate structural Solovev gate:

- tangent iterations: 23 -> 1
- adjoint iterations: 41 -> 2
- warm tangent: 3.51 ms -> 0.90 ms
- warm adjoint: 16.06 ms -> 2.46 ms
- warm custom VJP: 16.66 ms -> 2.37 ms
- one-time factor build: 3.02 s
- cold tangent/adjoint/custom VJP: 4.17 / 5.57 / 6.17 s

The factors are bounded by (|n|, m-band) and retain all radial and R/Z/lambda couplings inside each band. Peak compile RSS is recorded rather than described as live solve memory.

## Validation

- 35 polishing/preconditioner/implicit tests pass in 163.86 s
- focused primal/transpose block recovery and custom-VJP tests pass
- Ruff and git diff checks pass
- Sphinx strict dummy build passes

This remains stacked on draft #175. It closes the transpose-cost promotion gate there; block-local equation/coordinate equilibration and certified endpoint completion remain separate gates.


rogeriojorge:

PR #177, opened 2026-08-29
Cut QA optimization startup time and memory

rogeriojorge:
## Summary
- construct the fixed-boundary evolved-DOF mask analytically from exact mode-table invariants instead of rediscovering it with two force passes and reverse sweeps
- assemble raw radial blocks with bounded chunked VJPs, reducing first-use graph size and peak memory
- make `QA_optimization.py` a single max-mode scalar L-BFGS-B problem, using one reverse implicit solve for the exact scalarization of the existing weighted residual rows
- retain the optimized block least-squares lane for users who need pointwise Gauss-Newton residual geometry
- add a cache-disabled startup benchmark that measures distinct nearby decision vectors rather than identical-point memoization hits

## Corrected Apple M4 evidence
All records use 48 boundary DOFs, `max_mode=3`, JAX x64, and the persistent compilation cache disabled.

| lane | cold startup | peak RSS | warm distinct value+gradient |
|---|---:|---:|---:|
| `main` least-squares | 55.93 s | 3705.6 MiB | 15.02 s (one point) |
| branch least-squares | 44.73 s | 2965.1 MiB | 16.57 s (median of 3) |
| branch scalar example | 32.21 s | 2574.0 MiB | 17.38 s (median of 3) |

The scalar and least-squares initial costs agree to `2.8e-8` absolute and their gradient norms to about `2.4e-8` relative. The warm points are deterministic steps of `1e-5 * problem.scales`; they bypass `FunctionProblem`'s identical-vector memoization.

Artifacts:
- `benchmarks/qa_optimization_startup_main_m4.json`
- `benchmarks/qa_optimization_startup_least_squares_m4.json`
- `benchmarks/qa_optimization_startup_scalar_m4.json`

## Verification
- Ruff and benchmark JSON validation pass
- fixed-boundary analytic mask is tested against the independent numerical structural oracle
- chunked block assembly is tested against the dense reference and validates invalid chunk sizes
- scalar objective/gradient is tested against the least-squares scalarization
- the full QA example gate completed and required a strict optimizer cost decrease (`1 passed, 2 skipped` in 164.86 s on the earlier standalone commit)

## Claim boundary
This PR targets startup latency and memory. It does not claim that a scalar quasi-Newton step is faster than a block Gauss-Newton step: the corrected distinct-point measurements are comparable (17.38 s versus 16.57 s). The example chooses the scalar lane for its 12.5 s faster cold startup and roughly 391 MiB lower peak RSS; the pointwise least-squares API remains available and is also substantially cheaper to construct than on `main`.

### DIFF
--- a/benchmarks/qa_optimization_startup.py
+++ b/benchmarks/qa_optimization_startup.py
@@ -111,10 +111,17 @@ def main() -> None:
         evaluation = problem.compile_residual_and_jacobian(progress=False)
     compile_seconds = time.perf_counter() - started
 
+    # FunctionProblem intentionally memoizes an identical decision vector, so
+    # repeating x0 would measure a key lookup rather than the warm equilibrium
+    # plus implicit-derivative path used by an optimizer.  Use deterministic,
+    # distinct, in-bounds steps after compilation instead.
+    direction = np.linspace(-0.5, 0.5, problem.x0.size)
+    direction /= np.linalg.norm(direction)
     warm = []
-    for _ in range(args.repeats):
+    for repeat in range(args.repeats):
+        point = problem.x0 + (repeat + 1) * 1.0e-5 * problem.scales * direction
         started = time.perf_counter()
-        problem.value_and_grad(problem.x0)
+        problem.value_and_grad(point)
         warm.append(time.perf_counter() - started)
 
     report = {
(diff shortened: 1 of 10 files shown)


rogeriojorge:
The first hosted run exposed a real example-contract mismatch after replacing least-squares with the scalar L-BFGS-B lane: `tests/test_examples.py` only recognized least-squares table rows, so it could not verify cost decrease even though the example completed. Commit `11976097` now prints the exact scalar initial/final optimizer cost from the evaluations already performed and extends the shared gate to require a strict decrease for either output format. Local evidence on the standalone branch: Ruff passed; the source contract passed; the full QA example gate passed (`1 passed, 2 skipped`) in 164.86 s. No scientific tolerance or optimizer acceptance condition was relaxed.

PR #180, opened 2026-08-29
Weight strong roots by coordinate volume

rogeriojorge:
## Summary

- multiply both independent physical strong-force channels by `abs(sqrt(g))` before Fourier/radial projection, matching the DESC force objective
- preserve the off-axis zero set and leave the coordinate-gauge channel unchanged
- commit clean production-size conditioning, timing, and peak-memory evidence

## Measured production diagnostic

At `ns=11`, `mpol=6`, degree 5, 117 free coordinates on Apple M4:

- numerical rank at relative tolerance `1e-8`: 112 -> 113
- condition estimate: `8.07e10` -> `8.83e9`
- median warm residual/JVP: `3.01/4.60 ms`
- JVP finite-difference relative error: `5.80e-9`
- runtime-build peak RSS increase: `867 MiB`

Weighting only one force channel was rejected because it badly imbalances the singular spectrum. Adding tested spectral-condensation gauge weights raises rank to 114 but worsens the best condition estimate to `1.43e10`; those weights are diagnostic only.

## Production continuation result

The dependent startup fix in #181 removes 80 nonlinear / 1423 linear iterations spent chasing a `7.39e-13` alpha=0 remainder below an artificial `~9e-16` floor. The corrected run reached pseudo-arclength, but cold compilation exceeded 6.5 minutes and about 3.3 GiB RSS before the first bordered corrector completed. The run was stopped at the explicit performance gate.

This means the volume-weighted formulation improves the local spectrum but currently fails the cold-runtime gate; it is not a merge candidate. SOLVAX#94 and VMEX #181 remove repeated branch-data recompilation before the production continuation is reconsidered.

## Verification

- `tests/test_polish_preconditioner.py`: 41 passed on the stacked startup head
- `tests/test_performance_docs.py`: 7 passed
- Ruff, docs prose/media, JSON parse, and `git diff --check`: pass
- clean-commit benchmark: `benchmarks/strong_root_m6_volume_weighted_m4.json`

## Claim boundary

This is conditioning evidence, not an alpha=1 solution. Rank is still 113/117. The weakest right direction is dominated by axisymmetric `L_sin(m=1)`; the next three by `R_cos(m=0)`, with all four left vectors dominated by the highest retained `m=5` coordinate-gauge equations. The remaining defect is mixed physical/gauge coupling. The public polish/implicit APIs stay draft until the independent certificate and runtime gates pass. Full hosted CI is intentionally deferred while this remains an experiment.

### DIFF
--- a/benchmarks/strong_root_m6_volume_weighted_m4.json
+++ b/benchmarks/strong_root_m6_volume_weighted_m4.json
@@ -1,0 +1,56 @@
+{
+  "case": "solovev",
+  "cold_jvp_seconds": 0.9538727920007659,
+  "cold_residual_seconds": 0.37170395900466247,
+  "compilation_cache": "default",
+  "condition_number": 8828386081.32041,
+  "coordinate_scale_range": [
+    7.797769613049024e-07,
+    0.009102074687804574
+  ],
+  "degree": 5,
+  "equation_scale_range": [
+    3.847172085464133e-06,
+    2675.7953759327415
+  ],
+  "free_dofs": 117,
+  "initial_low_residual_norm": 7.387614345725676e-13,
+  "initial_scaled_residual_norm": 8.982800554888583e-08,
+  "input_data_embedded": false,
+  "jvp_peak_rss_increase_mib": 170.453125,
+  "jvp_relative_error": 5.803914906696761e-09,
+  "largest_singular_value": 9.303057871762458e-10,
+  "measurement_commit": "504073e49a6d9dbe91d1874daea03627717f230a",
+  "measurement_dirty": false,
+  "mpol": 6,
+  "ns": 11,
+  "operator_balance": 120415161.84510404,
+  "platform": "macOS-26.6.2-arm64-arm-64bit",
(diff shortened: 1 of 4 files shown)


rogeriojorge:
Follow-up Schur diagnostic (same `ns=11`, `mpol=6`, degree-5 volume-weighted operator): eliminating all 35 independent coordinate-gauge directions produces an 82x82 physical/lambda operator with numerical rank 79/82 at relative tolerance `1e-8` and condition estimate `9.63e9`. The unweighted reduced operator was only 73/82 at `3.74e11`. The weakest reduced pair is now `L_sin(m=1)` against the helical-force `m=1` row; the next two weak directions mix high-m geometry with radial-force rows. This confirms that volume weighting improves physical rank, not merely the gauge coupling, but also leaves three near-axis physical directions unresolved. The next experiment is an exact gauge-eliminated 82-coordinate root plus a near-axis chart; this PR remains draft and does not claim alpha=1.

rogeriojorge:

PR #178, opened 2026-08-29
Equilibrate strong roots with streaming JVPs

rogeriojorge:
## Summary
- equilibrate the strong-root row and coordinate spaces with deterministic streaming JVPs
- retain only O(n) live scaling data and never retain a production dense Jacobian
- propagate the invertible coordinate/row maps through the primal, continuation, tangent, adjoint, and custom-VJP paths
- preserve exact transpose algebra and avoid regularizing the unshifted mode-block inverse
- benchmark a fixed physical perturbation direction so finite differences remain meaningful after a coordinate change

## Production evidence
Paired clean-commit Apple M4 runs use `ns=11`, `mpol=6`, degree 5, 117 unknowns, 20 repeats, and `VMEX_COMPILATION_CACHE=disabled`:

| metric | #176 parent | streaming equilibration |
|---|---:|---:|
| numerical rank at 1e-8 | 39/117 | 112/117 |
| condition number | 1.40e17 | 8.07e10 |
| runtime construction | 56.1 s | 98.7 s |
| first residual | 3.25 s | 4.47 s |
| first JVP | 4.20 s | 6.26 s |
| warm residual | 7.07 ms | 10.05 ms |
| warm JVP | 10.43 ms | 17.60 ms |
| physical-direction FD/JVP error | 2.12e-8 | 6.22e-9 |

Artifacts: `benchmarks/strong_root_m6_unscaled_m4.json` and `benchmarks/strong_root_m6_equilibrated_m4.json`.

## Verification
- `python -m pytest tests/test_polish_preconditioner.py -q`: 36 passed
- focused low-inverse, transpose-duality, tangent/adjoint/custom-VJP, and bordered tests: 4 passed
- Ruff passes for all changed Python files
- `sphinx-build -W -b dummy docs docs/_build/dummy`: passed
- both benchmark artifacts are clean-commit and cache-disabled

## Claim boundary / draft gates
This is a material conditioning improvement, not a certified endpoint claim. Keep this PR draft until:

- the remaining five numerical directions are resolved or proven to be an intentional physical gauge at the production tolerance;
- the 42.6 s additional true-cold runtime construction cost is reduced or accepted with a documented amortization policy;
- continuation reaches and independently certifies `alpha=1` on the plan case;
- the stacked native-coordinate and mode-block PRs (#175 and #176) are accepted.

PR #181, opened 2026-08-29
Reuse branch-connected polish startup work

rogeriojorge:
## Summary

- accept the mathematically solved alpha=0 endpoint from its pre-equilibration RMS instead of driving roundoff through PTC
- pass tangent and predictor dynamically to one reusable SOLVAX bordered corrector
- retain the old PTC fallback for a genuinely inconsistent low endpoint and compatibility with older SOLVAX call shapes

## Production diagnosis

At `ns=11`, `mpol=6`, degree 5, the low endpoint residual was `7.39e-13`, but the old operator-balance floor demanded about `9e-16`. Both legacy and mode-block paths spent 80 nonlinear / 1423 linear iterations and failed at alpha=0.

After bypassing that roundoff chase, the volume-weighted diagnostic reached arclength, but cold compilation exceeded 6.5 minutes and about 3.3 GiB RSS. The dependent SOLVAX PR #94 makes branch data dynamic so later correctors reuse one compiled executable. Its scalar gate measured 0.278 s cold and 0.00012 s for a second corrector with different branch data; this is mechanism evidence, not an end-to-end VMEX speedup claim.

## Verification

- `tests/test_polish_preconditioner.py`: 41 passed in 143.67 s
- Ruff and `git diff --check`: pass
- normalized endpoint rescaling regression: pass
- parameterized bordered-preconditioner path: pass

## Dependencies and claim boundary

- stacked on draft #178
- repeated-compile path requires uwplasma/SOLVAX#94
- remains draft until the production alpha=1 certificate and runtime gate pass

### DIFF
--- a/docs/explanation/high-order-force-balance.rst
+++ b/docs/explanation/high-order-force-balance.rst
@@ -217,8 +217,12 @@ bounded number of JVP/low-inverse applications; it is stored as
 Branch-preserving driver
 ------------------------
 
-The fixed-boundary host driver first solves the ``alpha=0`` legacy consistency
-endpoint, then calls SOLVAX adaptive continuation with state-dependent JVPs,
+The fixed-boundary host driver first verifies the ``alpha=0`` legacy
+consistency endpoint before row equilibration.  Because that endpoint
+subtracts the stored legacy defect, the zero correction is its mathematical
+root; accepting its roundoff-level remainder avoids an unnecessary nonlinear
+solve below floating-point noise.  A genuinely inconsistent endpoint still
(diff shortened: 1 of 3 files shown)


rogeriojorge:
Dependency update: uwplasma/SOLVAX#94 merged as `85d09c47562eec5257f90bd2a7dedc8834292c05` after all platform, types, docs/build, combined branch-coverage, and Codecov gates passed. This VMEX draft now has a released-main dependency path for the cache-safe parameterized corrector; it remains draft because the stacked physical alpha=1 certificate/runtime gates are still open.

rogeriojorge:

PR #182, opened 2026-08-29
Diagnose an exact gauge-free strong root

rogeriojorge:


rogeriojorge:
The automatic full numerical matrix was canceled while this PR is a stacked diagnostic draft. Local changed-module tests and the production rank/JVP benchmark are recorded in the body. The full matrix, changed-line coverage, packaging/docs, and PR gate are required once the physical chart is full rank and this becomes a merge candidate.

rogeriojorge:
Follow-up rejected experiment: physical-disk-metric whitening of the gauge-free spline chart does not recover the three weak directions. With the same Ruiz/SVD gate, raw rank is 79; input-measure whitening gives 77; primal input/output whitening gives 78; dual-measure whitening gives 70 and worsens conditioning by orders of magnitude. This rules out mass-normalizing the existing analytic rho^m q(rho^2) span. It does not rule out a genuinely different Fourier-Zernike trial/test space; the next hypothesis must change the near-axis space or force asymptotics.

rogeriojorge:
Angular-convergence follow-up: rebuilding the same gauge-free diagnostic with 25 poloidal points raises rank 79 -> 80; 37 points remains rank 80. Thus one direction was an under-resolved nonlinear projection, the gain is converged, and two weak physical directions remain. Conditioning is not improved. This warrants a separate explicit angular-overintegration draft with runtime/memory evidence, not promotion of this PR.

rogeriojorge:

PR #183, opened 2026-08-29
Converge nonlinear angular force projection

rogeriojorge:


rogeriojorge:
The automatic full matrix was canceled while this remains a stacked diagnostic draft that explicitly fails rank, runtime, and memory gates. Focused structural tests and the clean production-resolution benchmark are recorded in the body. The complete matrix and changed-line coverage are required only if a later full-rank, performance-acceptable result makes this a merge candidate.

rogeriojorge:
Root cause found above this layer: `BSplineBasis.quadrature_nodes` are in normalized flux s, while `evaluate_strong_force` accepts rho=sqrt(s). The root passed s directly as rho but fit against B(s). Recomputing consistently at rho=sqrt(s_quadrature) raises the 25-point gauge-free diagnostic from rank 80/82 to 82/82 and improves the equilibrated condition estimate from ~5.04e9 to ~6.63e7. This will be isolated in a child PR with a coordinate regression and full continuation/performance gates.

rogeriojorge:

PR #197, opened 2026-08-30
Add scalar-adjoint QA/QH/QP/QI optimization examples

rogeriojorge:
## Summary

- restore `examples/optimization/QA_optimization.py` to the staged SciPy `least_squares` residual/Jacobian workflow
- add separate scalar-adjoint QA, QH, QP, and QI drivers in both vacuum and finite-beta form
- document exactly why a scalar objective needs one equilibrium-adjoint solve per gradient, and when that is preferable to a residual Jacobian
- keep the implementation on a branch cut directly from `main`; this PR is independent of and does not modify #196

## One adjoint for one scalar gradient

Let `z` denote the converged equilibrium, `x` the design variables, and

```text
g(z, x) = 0,             Phi(z, x) = 1/2 r(z, x)^T r(z, x).
```

The explicit least-squares interface supplies every residual and the full residual Jacobian,

```text
J_r = d r / d x = r_x - r_z g_z^{-1} g_x,
grad_x Phi = J_r^T r.
```

This is useful because SciPy can build a Gauss-Newton trust-region model and retain residual-level diagnostics, but VMEX must compute and materialize the `n_residuals x n_dofs` Jacobian.

The scalar interface forms `Phi` before differentiating. Its implicit reverse rule solves

```text
g_z^T lambda = Phi_z^T = r_z^T r,
grad_x Phi = Phi_x - lambda^T g_x.
```

That is one equilibrium-adjoint right-hand side for each scalar gradient evaluation, independent of the residual-row and design-variable counts. It does not mean constant runtime: equilibrium resolution, linear conditioning, and the objective VJP still determine the work. The documentation also calls out that L-BFGS-B and nonlinear least squares have different curvature models, iterates, stopping rules, and potentially local basins.

## Example organization

The canonical QA file is again the explicit nonlinear least-squares tutorial. The new files are additive:

| Symmetry | Vacuum | Finite beta |
| --- | --- | --- |
| QA | `QA_optimization_scalar.py` | `QA_optimization_finite_beta_scalar.py` |
| QH | `QH_optimization_scalar.py` | `QH_optimization_finite_beta_scalar.py` |
| QP | `QP_optimization_scalar.py` | `QP_optimization_finite_beta_scalar.py` |
| QI | `QI_optimization_scalar.py` | `QI_optimization_finite_beta_scalar.py` |

Each runnable file keeps its physics, targets, continuation schedule, final solve, output names, and validation visible. A 76-line `_scalar_driver.py` centralizes only the repeated L-BFGS-B/scaled-coordinate wiring. The finite-beta cases calibrate a prescribed linear pressure profile and include radially weighted Mercier and resistive-interchange residuals. Final force-balance polishing remains an explicit, disabled-by-default option.

No VMEX source change was necessary: `VmecProblem.from_loss`, `residuals_from_tuples`, and `compile_value_and_gradient` already expose the required API.

## Performance evidence and limits

The reference page links the committed QA startup records rather than adding another benchmark variant. For 6,723 rows and 48 boundary degrees of freedom on the recorded Apple CPU host:

| Lane | Cold startup | Peak RSS | Warm value/gradient median |
| --- | ---: | ---: | ---: |
| residual Jacobian | 44.7 s | 2,965 MiB | 16.6 s |
| scalar adjoint | 32.2 s | 2,574 MiB | 17.4 s |

This supports the cold-start and memory motivation without claiming a universal warm-runtime win. Scalar objectives must be JAX-traceable; host-only callbacks still require a traceable implementation or finite differences.

## Validation

- `ruff check examples tests/test_examples.py`
- Python byte-compilation for all nine new modules
- `tools/check_docs_prose.py`
- Sphinx documentation build with warnings treated as errors (52 sources)
- focused scalar API and example-contract tests: 12 passed
- restored canonical QA end-to-end CI smoke: passed
- vacuum and finite-beta QA scalar pytest smoke tests: passed
- direct CI-smoke execution of all eight scalar examples: passed, with finite initial/final costs and distinct input, WOUT, monitor, and plot outputs

Observed CI-smoke scalar cost changes:

```text
QA vacuum       21.4488 -> 0.4908
QA finite beta  22.6159 -> 21.4269
QH vacuum        0.8144 -> 0.4523
QH finite beta   0.3506 -> 0.3325
QP vacuum       10.9987 -> 1.1926
QP finite beta  12.7535 -> 7.3061
QI vacuum        7.3408 -> 0.6868
QI finite beta   5.2069 -> 0.7267
```

## Review focus

1. Is retaining `least_squares` as the canonical QA tutorial while adding scalar companions the right API teaching split?
2. Are the optimizer differences and the meaning of “one adjoint” stated precisely enough?
3. Are the finite-beta objective weights suitable as concise starting examples rather than production configurations?

### DIFF
--- a/docs/reference/optimization.rst
+++ b/docs/reference/optimization.rst
@@ -120,6 +120,12 @@ steps. Their iterates, stopping tests, and possibly the local minimum reached
 can differ. Always compare final physical terms and held-out validation
 metrics, not iteration counts alone.
 
+Measured on the QA workflow at a matched evaluation budget, the least-squares
+driver reached roughly a 3x lower objective than the scalar lane; the scalar
+lane's gains are a cheaper cold start and lower peak memory (44.7 s to 32.2 s
+wall, 2965 to 2574 MiB peak RSS on an Apple M4), so it is the cold-start and
+low-memory option, not a replacement for the least-squares driver.
+
 Choose deliberately:
 
 * Keep the vector route when residual-level diagnostics, a least-squares trust
--- a/docs/reference/optimization.rst
+++ b/docs/reference/optimization.rst
@@ -42,6 +42,110 @@ profiles, and transforms (it is not elapsed run time). Use
 :meth:`~vmex.core.problem.FunctionProblem.from_functions` when the user already
 has decision-vector-level functions and derivatives.
 
+Vector least squares or one scalar adjoint
+-------------------------------------------
+
+The residual and scalar interfaces can represent the same mathematical cost
+while asking for different derivatives. Let ``z`` be the converged equilibrium,
+``x`` the boundary/current decision vector, and
+
+.. math::
+
+   g(z,x)=0, \qquad
+   \Phi(z,x)=\tfrac12 r(z,x)^T r(z,x).
+
+The residual/Jacobian route builds
+
+.. math::
+
+   J_r = \frac{d r}{d x}
+       = r_x-r_z g_z^{-1}g_x,
+   \qquad \nabla_x\Phi=J_r^T r.
+
+This is what ``VmecProblem.from_tuples`` plus SciPy ``least_squares`` needs.
+The optimizer receives every residual row and the complete
+``n_residuals``-by-``n_dofs`` Jacobian. It can form a Gauss--Newton trust-region
+model, report individual rows, and stop on residual/Jacobian criteria. VMEX
+uses its block response algorithm to amortize the equilibrium linear algebra,
+but the complete residual Jacobian is still computed and materialized.
+
+When the optimizer only needs a scalar value and gradient, form ``Phi`` before
+differentiating. The adjoint is
+
+.. math::
+
+   g_z^T\lambda=\Phi_z^T= r_z^T r,
+   \qquad
+   \nabla_x\Phi=\Phi_x-\lambda^T g_x.
+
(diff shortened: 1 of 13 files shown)


rogeriojorge:
Thanks for this, and sorry for the slow verdict — I wanted to measure it rather than guess.

Verdict: I would like to take these in as examples, framed as the cold-start and low-memory option, rather than as a replacement for the least-squares driver.

What I measured, on a rebase of this branch onto current main (main has moved a long way since it was opened):

- The two latencies that motivated the scalar lane — the wait before `dof_names` appears, and the gap after "Value and gradient ready" — are not specific to the least-squares path. They came from shared startup machinery, and #240 fixed them there: cold time to first output on the QA workflow went from 62 s to 6.5 s. The scalar lane does not change them.
- Cold start and peak memory do improve on the scalar lane: 44.7 s -> 32.2 s wall, 2965 -> 2574 MiB peak RSS.
- At a matched evaluation budget, trust-region-reflective least squares reached roughly a 3x lower objective than the scalar optimizer on the same problem. For objective progress per evaluation the existing driver still wins.

So the value here is real but it is a different trade, and the examples should say so. Before merge I would ask for: a rebase onto main, and one sentence in each example stating that the scalar lane trades objective progress per evaluation for a cheaper cold start and lower peak memory. I already have a clean rebase locally if it is easier for me to push it — say the word and I will.

PR #213, opened 2026-08-31
Improve implicit refinement reuse and expose optimization controls

csvega24:
## Summary

This PR improves repeated implicit-gradient evaluations in optimization workflows and exposes related public controls.

- Add `VmecProblem.jax_quantity_from_state()` for evaluating differentiable quantities directly from the accepted implicit equilibrium state.
- Expose virtual-casing `chunk_size` and `target_chunk_size` through the VMEX exterior-field APIs.
- Add guarded reuse of fixed-point-refinement displacements between nearby optimization points.
- Expose the fixed-point refinement tolerance through `refine_tol`.
- Add diagnostic instrumentation and regression benchmarks for implicit solves and refinement.

## Motivation

VMEX refines each converged host equilibrium before evaluating implicit derivatives. In optimization workflows, this refinement can account for a substantial fraction of the evaluation time, even when consecutive trial points are nearby.

This PR reuses the previous accepted refinement displacement as an initial guess at the next point. The candidate is accepted only when it reduces the exact frozen residual and reaches the configured refinement tolerance. Otherwise, VMEX discards it and executes the original refinement path.

### Differentiable state quantities

```python
quantity, status = problem.jax_quantity_from_state(x, quantity_fn)
```

This evaluates a fixed-shape differentiable quantity using the same accepted equilibrium state as the implicit objective. It avoids independently materializing an equilibrium for related surface quantities and allows callers to form VJPs with standard JAX operations.

### Virtual-casing chunking

`chunk_size` and `target_chunk_size` can now be passed through `VmecProblem.exterior_field()` and `VmecExtender`. These settings control the memory/throughput tradeoff in virtual-casing field evaluations.

### Refinement tolerance

```python
problem = VmecProblem.from_tuples(..., refine_tol=1.0e-10)
```

The default remains `1.0e-10`. Setting `refine_tol=numpy.inf` disables post-solve refinement and uses the converged host state directly.

## Correctness And Fallback Behavior

The reuse mechanism does not alter the equilibrium equations or requested refinement tolerance.

- Reused displacements are initial guesses only.
- A candidate must reduce the exact frozen residual.
- A reused result is cached only after satisfying `refine_tol`.
- Failed or incompatible solves invalidate the reusable displacement.
- If reuse does not reach tolerance, VMEX reruns the original refinement path.

The strict default therefore preserves the existing refined-state derivative contract.

## Performance

On a representative 48-surface-DOF vacuum optimization benchmark, guarded reuse reduced median residual-evaluation time:

- from approximately `14.95 s` to `6.50 s` in one evaluation ordering;
- from approximately `10.12 s` to `8.09 s` in the reverse ordering.

The final residuals and gradients remained within the existing accuracy gates.

Production-style vacuum and finite-beta replays also showed reduced refinement work without a significant memory regression. A relaxed `refine_tol=1.0e-8` was validated as an optional throughput setting, but the strict `1.0e-10` tolerance remains the default.

## Validation

Tests cover:

- exact and nearby-point reuse;
- candidate acceptance and rejection;
- fallback to the original refinement path;
- failed-solve and configuration-change invalidation;
- gradient and directional finite-difference agreement;
- virtual-casing chunk-control forwarding;
- differentiability of `jax_quantity_from_state()`;
- vacuum and finite-beta performance replays.


rogeriojorge:
Superseded by three independent PRs rebased on current main:\n\n- #217: guarded implicit-refinement reuse and the public refinement tolerance, with the performance/accuracy evidence and 95% Codecov wiring\n- #216: differentiable state-quantity composition API\n- #218: virtual-casing source/target chunk controls\n\nThe split preserves the useful production ideas from this PR while removing the twelve Phase 0-5 benchmark/summarizer files, private diagnostic-hook machinery, machine-specific replay assumptions, and unrelated coupling between APIs. Each replacement is conflict-free, documented, and carries focused numerical/physics tests. This PR is closed only because its work is now represented by those reviewable replacements.

PR #217, opened 2026-08-31
Reuse guarded implicit refinement corrections

rogeriojorge:
## Motivation

VMEX Newton-refines the host equilibrium before implicit differentiation because the theorem applies at the frozen residual root, not merely at the host solver's sum-of-squares stopping point. Nearby optimization trials repeatedly solve almost the same refinement problem. #213 demonstrated useful warm-runtime savings, but mixed this algorithm with two public APIs, private instrumentation, and twelve benchmark scripts.

This PR keeps only the refinement algorithm and its user-facing accuracy control.

## Algorithm and numerical guard

For a nearby parameter point, VMEX proposes the preceding accepted refinement displacement as the Newton initial guess. It evaluates the new frozen residual exactly before using it. A candidate is used only when it lowers that residual. If the warm path does not reach `refine_tol`, VMEX discards it and replays the original path from the host-solver state. Rejected equilibrium trials invalidate the correction.

Thus the optimization changes work, not the accepted fallback result. The strict default remains `refine_tol=1e-10`; `numpy.inf` is an explicit legacy-comparison switch. Nonpositive, NaN, and negative-infinite tolerances are rejected.

## Measured evidence

The pre-split review reproduced #213 on CPU with 48 active boundary degrees of freedom and 41,475 residual rows. Against the legacy refinement path:

| evaluation | legacy | guarded reuse | change |
|---|---:|---:|---:|
| residual, plus perturbation | 12.30 s | 7.56 s | -39% |
| residual, minus perturbation | 4.48 s | 2.45 s | -45% |
| scalar value + gradient, plus | 17.58 s | 12.97 s | -26% |
| scalar value + gradient, minus | 9.64 s | 7.11 s | -26% |

Cold problem construction was unchanged (23.01 vs 23.06 s), as expected: this targets repeated nearby refinement. Objective differences were below 3.3e-12; gradient relative L2 difference was at most 1.04e-8; both paths reached `refine_tol=1e-10`. Reverse-order measurements retained an overall improvement but showed the expected compilation/order noise at one point.

## Coverage and repository hygiene

- Remove the instrumentation and all twelve Phase 0-5 benchmark/summarizer files proposed in #213; no new benchmark file or folder is added.
- Test the residual-improvement guard, exact warm acceptance, and legacy fallback in one focused numerical test; retain the existing real LASYM fixed-root and frozen-path FD tests as the physics anchors.
- Publish the already-combined CI coverage XML to Codecov with OIDC.
- Set both project and patch Codecov targets to 95% and add the coverage badge to the README. The existing local diff-coverage gate remains 95%.

## Verification

- `ruff check vmex tests`
- `mypy vmex`
- Focused refinement and public-factory tests: 2 passed
- `sphinx -W -j auto -b html docs docs/_build/html`
- Earlier full implicit-gradient campaign on the submitted algorithm: 6 full physics tests passed, including the real 3-D LASYM fixed-root and frozen-path gradient checks.

Supersedes the refinement/tolerance portion of #213; independent of its state-quantity and virtual-casing API portions.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -308,6 +308,9 @@ jobs:
     runs-on: ubuntu-latest
     timeout-minutes: 5
     needs: [changes, fast, physics, parity, device]
+    permissions:
+      contents: read
+      id-token: write
     if: >-
       always() &&
       needs.changes.outputs.run_coverage == 'true' &&
@@ -333,6 +336,12 @@ jobs:
           coverage combine coverage.device coverage.fast coverage.physics-* coverage.parity-*
           coverage report --include='*/vmex/mirror/*' --fail-under=90
           coverage xml
+      - name: Publish combined coverage
+        uses: codecov/codecov-action@v5.5.5
+        with:
+          fail_ci_if_error: true
+          files: coverage.xml
+          use_oidc: true
       - name: Enforce changed-line coverage
         if: github.event_name == 'pull_request'
         run: diff-cover coverage.xml --compare-branch="origin/${GITHUB_BASE_REF}" --fail-under=95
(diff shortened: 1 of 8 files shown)


rogeriojorge:
Current-branch physics follow-up completed after opening the PR: `RUN_FULL=1 pytest -q tests/test_implicit_grad.py::test_lasym_3d_state_is_anchored_at_the_frozen_root tests/test_optimize.py::test_least_squares_implicit_warm_start_modes` — **2 passed in 326.31 s**. This verifies both the real 3-D LASYM frozen-root anchor and equality of the optimization result under the established plain-hot versus perturbation warm-start policies on commit 5211837d.

rogeriojorge:
Memory note from the same 48-DOF / 41,475-row replay: sampled peak RSS did not increase relative to the legacy refinement path. The cache is deliberately bounded to one correction pytree per live canonical implicit configuration; rejected trials remove it. No history, residual vectors, Jacobians, or benchmark telemetry are retained.

PR #220, opened 2026-09-01
Repair the strong-force guards against the honest v0.8.0 evidence

rogeriojorge:
## Summary

On main, `tests/test_performance_docs.py` has been red on a pristine checkout since 639aec7c (#204): `test_solvax_polish_artifact_is_independently_certified` and `test_readme_strong_force_figure_matches_committed_sources` both die with `TypeError` on nulls the regeneration wrote into the committed artifacts. Both guards predate #204 (they were added with v0.8.0 in b0e7efda), so the regeneration went out from under them. The verdict here is a split: one null was lost-but-still-valid evidence that belongs back in the artifact; the rest were guards stale against #204's own documented, honest re-measurement.

## Changes

- **Restore the summary evidence block** in `benchmarks/strong_force_comparison_m4.json`. #204's regeneration omitted `--before-summary/--after-summary`, so the generator nulled the block — but the committed `readme_polish_summary.webp` stayed byte-identical to what v0.8.0 certified (its sha256 still matches the restored value) and the README kept displaying it. The restored values describe the exact committed bytes; nothing is re-measured or fabricated.
- **State the summary contract in the guard** instead of raising `TypeError`: a null summary block is legal only while the README does not display the summary figure; a displayed figure must match its committed evidence.
- **Accept release-pinned solvax provenance**: the #204 measurement ran solvax from a non-checkout install, so `strong_polish.py` correctly recorded `solvax_source: null` while pinning `versions.solvax = "0.20.0"` — the same code the old artifact certified as commit `5a49926` (the release/0.20.0 merge). The guard now requires either clean git provenance or a pinned release version.
- **Treat `least_squares_success` as the diagnostic #204 declared it to be**: the acceptance contract is the independent certificate alone (volume-L2 bar, radial-refinement stability, positive signed Jacobian — all still asserted), and the solver flag goes False when the step budget expires on an already-certified state.
- **Update the stale timing bounds** to the honest two-regime measurements: tokamak first run < 150 s plus persistent-cache rerun < 60 s (measured 103.4 s / 30.0 s), and for the QA case the first run must beat DESC outright while the rerun beats it tenfold (19.5 s and 3.6 s vs 153.1 s) — matching the README's actual claim.

Deliberately **not** done here: re-running the benchmarks. There is no committed assembler for `strong_force_cases_m4.json`, a rerun would cascade into new timings, a re-rendered figure, and stale README prose — and the figure policy set in 69cf1e3a (#212) calls for the runtime panels to leave the README figure entirely, which is its own work item.

## Test plan

- `python -m pytest -q tests/test_performance_docs.py` — 11 passed (was 2 failed, 9 passed on main).
- Verified `benchmarks/strong_force_comparison_m4.json` still round-trips as the generator's canonical `json.dumps(..., indent=2, sort_keys=True)` output.
- Verified the committed summary figure's sha256 matches the restored metadata, and that no other tests consume the touched fields.

### DIFF
--- a/benchmarks/strong_force_comparison_m4.json
+++ b/benchmarks/strong_force_comparison_m4.json
@@ -52,8 +52,11 @@
   "figure": "docs/_static/figures/readme_strong_force_comparison.webp",
   "figure_sha256": "39e4ce0f6b39bd360c1c46fec0b5dbbf526dcded4cd3c490167f58bd0bc2bca3",
   "schema": "vmex.strong-force-readme-figure/4",
-  "summary_figure": null,
-  "summary_figure_sha256": null,
-  "summary_independent_l2": null,
+  "summary_figure": "docs/_static/figures/readme_polish_summary.webp",
+  "summary_figure_sha256": "6c0f892e28c469b2937dd4e9a567c15cca8a0a4d54cbc575ab1ecc953d23568c",
+  "summary_independent_l2": {
+    "after": 0.005616607909342132,
+    "before": 0.013607094167290654
+  },
   "timing_note": "First-run wall time: process start, load, solve, and export on one Apple host with an empty JAX compilation cache; the top row is the bundled input.shaped_tokamak_pressure_polished deck. The VMEX rerun marker is the same command with the persistent cache the first run populated (on by default); the legacy codes have no JIT stage, so their reruns match their first runs."
 }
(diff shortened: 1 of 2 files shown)


rogeriojorge:
Superseded by #226, which consolidates this diagnosis (correct: the regeneration nulled the summary evidence out from under the guards) with #221's schema/policy repairs, adapted to the artifacts #215 put on main.

PR #221, opened 2026-09-01
Make the accuracy-only README figure pass its own guards

rogeriojorge:
Implements the plan's README figure policy (section 21, decided 2026-08-31): the comparison figure shows force-balance accuracy only — the first-run wall-time panels and rerun markers come out (two panels per row: radial profile and volume-L2 bars), the README caption loses its timing prose, and runtime evidence points to `benchmarks/baselines/` and PR records until VMEX wall times are competitive with the legacy codes.

**Relationship to #215:** that PR removes the same panels and goes further (new QA deck, row replacement in progress), but leaves `tests/test_performance_docs.py` untouched, so the guard suite still fails there (metadata schema `/4`, `summary_figure: null`, `timing_note` present). This PR is the narrower change that makes the guards and the committed artifacts agree today. If #215 lands first, this can be rebased down to just the evidence/guard repairs; if this lands first, #215 rebases to just its row-replacement work.

**Guard/artifact repairs** (disagreements left by the #204 regeneration, previously masked because the missing summary block aborted the figure test early with a `TypeError`):

- The metadata's summary-figure evidence block is restored: the committed `readme_polish_summary.webp` is byte-identical (sha `6c0f892e…`) to the version its recorded `1.361e-2 → 5.617e-3` independent errors were measured against. A plain re-render now carries the evidence forward and refuses when the on-disk summary no longer matches the recorded hash.
- Metadata schema bumps to `vmex.strong-force-readme-figure/5`: `timing_note` is replaced by a `figure_note` stating the accuracy-only policy; the test enforces both, and bans the old timing caption phrases from the README.
- SOLVAX provenance accepts a released wheel (`versions.solvax`, here 0.20.0) as well as a clean git checkout — the old guard hard-required a checkout.
- Acceptance follows the polish contract stated in `vmex/core/polish_driver.py`: `termination_reason == "independently-certified"`; the inner Gauss–Newton success flag is a diagnostic.
- The stale 90 s first-run bound now reflects the honest v0.8.0 evidence (103.4 s cold from an empty JAX cache, 30.0 s persistent-cache rerun, newly bounded at 60 s), and the DESC>10×-VMEX runtime gate leaves the README-figure test per the policy (the committed ratio is ~7.8×).

**Verification**

- `python -m pytest -q tests/test_performance_docs.py` — 11 passed (fails on `main` today).
- `python -m ruff check benchmarks/make_strong_force_comparison.py tests/test_performance_docs.py` — clean.
- A no-argument `python benchmarks/make_strong_force_comparison.py` reproduces the committed figure and metadata byte-for-byte.


rogeriojorge:
Superseded by #226: #215 landed the panel removal this PR carried, and #226 adapts the guard repairs to the post-#215 artifacts.

PR #271, opened 2026-09-05
Unify the flux-tube epsilon and R0 across the core and mirror lanes

rogeriojorge:
## Summary

Both gyrokinetic geometry exports set the GKX/GS2 contract key `epsilon` to `std(|B|)/mean(|B|)` and `R0` to `L_ref`. Verified against the installed GKX source: GKX reads `epsilon` as the inverse aspect ratio (`geometry/analytic.py:251`, `bmag = 1/(1 + eps cos theta)`) and derives `aminor = epsilon * R0` and `a_ref` when it writes run artifacts (`artifacts/nonlinear_netcdf.py:119,181`; read back as `epsilon = aminor/rmaj` in `artifacts/run_summary.py:159` and `geometry/flux_tube.py:452`). With `R0 = L_ref` that derived minor radius was meaningless in both lanes. #263 documented the mismatch on the mirror lane and deliberately left the value alone so the two lanes would not split; this PR picks one definition and applies it to both in the same change.

**Stacked on #263** (`audit/mirror-31-4`): the base is that branch so the diff shows only this change; GitHub retargets it to `main` once #263 merges.

## The definition

- `epsilon = (max|B| - min|B|) / (max|B| + min|B|)` along the sampled tube, as the new `vmex.core.turbulence.b_modulation_depth`, used by both `gk_fieldline_geometry` and `gk_closed_fieldline_geometry`. It equals GKX's `epsilon` exactly for its own `1/(1 + eps cos theta)` model and the local `r/R0` of any `1/R` tokamak field, and, unlike an aspect ratio, it exists on a straight mirror. The field-line mirror ratio is `(1 + eps)/(1 - eps)`. Hard max/min, smooth almost everywhere, the same depth `vmex.core.optimize.mirror_ratio` takes over a surface. The alternative, the geometric `Aminor_p/Rmajor_p`, is undefined on the mirror and is not what a stellarator field line sees.
- `R0` stops being `L_ref`: with `aminor = epsilon * R0` on the GKX side, `R0` has to be a major radius. Core lane: `R0 = volume_p / (2 pi <area>)`, the wout `Rmajor_p`, computed in the ballooning context from the same boundary quadrature as `L_ref` (live route) and read from `wout.Rmajor_p` with the same validation as `Aminor_p` (WOUT route; the existing live/WOUT parity test covers it at 2e-11). Mirror lane: `R0 = L_axis / (2 pi)`, which under its `L_ref = sqrt(V / (pi L_axis))` is the same identity `V / (2 pi^2 L_ref^2)`. `R_major` is also exported under `vmex` / `vmex_mirror` next to `L_ref` and `B_ref`.
- Neither scalar enters GKX's solver (only `analytic.py`, the artifact writers, the snapshot plot, and the `vmec_state_sensitivity` observable consume them), so no proxy value changes.

## Pinned-value audit

- No benchmark artifact or golden file pins the old value: the `"epsilon"` entries in `benchmarks/baselines/m4/C1_*.json` are the eps_eff timings, `tests/test_bootstrap.py`'s are the bootstrap geometry table, and `tests/test_wout_golden.py` pins `Rmajor_p`, which is unchanged.
- `tests/test_turbulence.py` checked key presence and live/WOUT parity only. It now also pins `epsilon` to the depth and `R0` to `wout.Rmajor_p`, checks `b_modulation_depth` against GKX's analytic model, and checks `epsilon ≈ sqrt(s) L_ref / R0` (rel 0.1) on the circular vacuum tokamak, which `std/mean` (about `eps/sqrt(2)`) fails.
- `tests/mirror/test_turbulence.py` (from #263) pinned `std/mean`; it now pins the depth, `epsilon == field_line_b_modulation`, and `R0 == axis_arc_length / (2 pi)`.
- GKX side: its bridge test is stubbed and does not depend on the values. Its `geometry/booz_xform_bridge.py:455` still emits the `std/mean` form under the same key, which is a GKX follow-up.

## Docs

`vmex.core.turbulence` module notes, the `gk_closed_fieldline_geometry` docstring, `docs/explanation/mirror-gyrokinetics.rst` (new definitions block after the normalizations), `docs/explanation/mirror-geometry.rst` (the "not mirror ratios" paragraph), the plan ledger entry 31.4-R3, and the CHANGELOG.

## Tests

```
tests/mirror/test_turbulence.py   3 passed
tests/test_turbulence.py          17 passed, 2 skipped (pre-existing gkx eigenvector-floor gates)
ruff                              clean
sphinx -b dummy                   0 warnings
```

## Measured before / after

Same lines the tests use (`s_index=7`; shaped tokamak at `alpha=0.3`, `ntheta=128`); "before" is the pre-PR code on the same states.

| case | `epsilon` before (std/mean) | `epsilon` after (depth) | `sqrt(s) L_ref / R0` | `R0` before (= `L_ref`) | `R0` after | wout `Rmajor_p` |
|---|---|---|---|---|---|---|
| shaped finite-beta tokamak | 0.1744 | 0.2705 | 0.2890 | 2.299 m | 6.075 m | 6.075 m |
| circular vacuum tokamak (`R = 6`, `a = 2`) | 0.1560 | 0.2176 | 0.2205 | 2.000 m | 6.000 m | 6.000 m |
| closed racetrack mirror (test fixture) | 0.2429 | 0.2802 | n/a | 0.346 m | 3.100 m (`L_axis / 2 pi`) | n/a |

On the circular tokamak the depth sits 1.3% from the local `r / R0`, where `std/mean` was 29% low (it is about `eps / sqrt(2)` for a cosine-like modulation). On the mirror the depth is `(R_m - 1) / (R_m + 1)` for the field-line `R_m = 1.7785` exported next to it.


rogeriojorge:
Reopened as #281 on main: GitHub closed this one when the merged base branch was deleted and will not reopen a PR whose base is gone.

PR #274, opened 2026-09-05
Revise the research plan after a second literature review

rogeriojorge:
Stacked on #270 (the tip of the #267 → #268 → #269 → #270 stack) so the logbook edits merge cleanly; retarget to `main` once that stack lands.

Three read-only literature reports were run against the plan of record (#267): the polishing method versus DESC, GVEC, SIESTA, SPEC and Thun et al. 2026 plus the least-squares literature; differentiable optimization versus DESC Parts 2–3, simsopt and the adjoint papers; and validation evidence versus every equilibrium-code paper of 2023–2026. Every claim below was checked against the code or a verified source; anything the reports could not verify is not in the plan.

## What changes in the plan

**Critical path.** `A → B → C0 → (C | F) → D`, with **E off the path**. C (recover 3-D polish) was a research bet with gates but no fallback; the new **C0** gives it three bounded experiments in order — a resolution-floor scan of the W7-X deck through VMEC++ (minutes), an LSMR swap on the tokamak (minutes), and an exact dense Gauss–Newton on a reduced W7-X to measure κ(J) and the chart's floor (hours) — a one-month budget, a kill criterion, and a fallback that is already publishable. E (single-problem sharding) had no demonstrated payoff anywhere in the field and leaves every paper's dependency list except a distributed-claim one. F and the first paper no longer wait on C.

**Why C0 says what it says.** The committed tokamak record is 80 Gauss–Newton steps and **47 308 CG iterations on a 148-unknown normal system** — ~590 per step where exact arithmetic needs 148 — the finite-precision signature of a badly conditioned normal matrix, before any 3-D measurement. CG on JᵀJ converges at a rate set by κ(J)², reaching `linear_rtol` 1e-3 needs ≈ 3.8 κ(J) iterations, and κ(J) of 10³–10⁴ is the natural expectation for a second-order differential operator (Hirshman–Betancourt). No equilibrium code in the field runs an unpreconditioned Krylov solver on that operator. And there is a resolution floor no fixed-chart polish can cross: DESC's own W7-X sits at 4.8e-2 / 3.8e-2 at M = N = 10 and Thun et al. 2026 at ~1e-2 — the W7-X 1.1% is angular truncation with a chart that can only move horizontally, not primarily a solver failure.

**C3.** LSMR on `[J; √μ I]` first (SOLVAX ships no LSQR/LSMR — that is the addition), then a same-basis block preconditioner (block-diagonal in (m, n), banded in the spline index, from the Hirshman–Betancourt terms), applied as J M⁻¹ with SPD companion (MᵀM)⁻¹. The #261 negative result for "B Bᵀ from the low-order inverse" is recorded as applying to that form only. Unfreeze Z_sin; add quadrature weights so the polish minimizes the certificate's norm.

**F0 / F2.** A committed derivative-verification table (multi-step centered FD from independent cold re-solves, contraction factor, duality, adjoint residual, resolution sweep) is what Antonsen 2019, Paul 2020, Nies 2022, Jorge 2023 and simsopt's test suite treat as the evidence; **VMEX has no Taylor test anywhere** (verified: zero mentions in `tests/`). DESC publishes none of this, so the table is a differentiator. Oracles that exist today and cost nothing to pin are listed (simsopt QS `total()` on the public LP-QA wout, DESC Γ_c on li383, NEO on W7-X via DESC's public file, simsopt `vacuum_well`, ConStellaration). The flagship optimization is a reproduction of Landreman–Paul precise QA, not a new design. The literature is unanimous with #266: no published free-boundary single-stage run starts from circular coils; the seeding chain and the Henneberg 2021 small-eigenvalue argument for the adjoint stagnation are written in.

**J2.** The four evidence categories every peer-reviewed paper presents (named-config comparison with a defined metric; error vs resolution with a rate; timing on a named machine with the compile policy; commit + data archive) and VMEX's gaps in reviewer order: the force figure that does not compare codes (#260: DESC native 4.0e-6 vs lifted 7.1e-2), claims main still makes that #256 withdraws, no convergence-with-resolution study of VMEX *solves* (the Panici Table 2 scans are cheap), timing methodology (five repetitions, serial-vs-serial, compiler flags), parity breadth (VMEC++'s 219-variant matrix), one Hudson-2025-type third-party anchor, packaging. Two competitor facts for positioning: VMEC++ 0.6.0 ships LASYM and 0.7.2 ships adjoint boundary sensitivities; DESC's anisotropic pressure has code but no paper.

**Sections 6–8.** Merge order for the seventeen open PRs with the edit each still needs; the next ten PRs; the environment runbook the migration dropped.

## Corrections made along the way (separate PRs where code moved)

- #272: the docs said Velasco et al. 2021 eqs. 20–21 relate prompt losses "approximately linearly" to Γ_c — a statement #250 introduced on my brief. Section 5.4 of the paper says the opposite for Γ_c; the three sites are corrected.
- #258: the `|Jacobian|` docutils substitution that failed `-W` is fixed; #254: the prune's failure branches are now covered (diff-cover 95%).
- #261: its parity lane reads an artifact that was never produced; the three-arm W7-X memory measurement is being regenerated on the office box from the branch head (flat arm 34.3 GiB, auto arm 3.0 GiB so far) and will be committed to that PR.
- The `physics/force-error-normalizations` branch turned out to hold four pushed commits (the non-saturating normalizations and the withdrawal of the 26-fold claim), not an uncommitted edit; it is being verified and gets a PR.

Plan-only plus the review manifest note; no production code changes.


rogeriojorge:
Superseded by https://github.com/uwplasma/vmex/pull/283, which keeps this PR's surviving ideas (bounded decision experiment, derivative table, evidence bar) in corrected form and answers #282's corrections. Closing so the reviewer compares two live proposals.

PR #275, opened 2026-09-05
Report force-balance normalizations that cannot saturate; withdraw the 26-fold claim

rogeriojorge:
Closes plan findings **31.2-R1**, **31.2-R3** and **31.2-R6**. The implementing agent pushed these four commits and was terminated before verifying them; verified today from its worktree on current main: `tools/preflight.py --static` PASS and `tests/test_strong_force.py` 33 passed (617 s).

**31.2-R1.** The certificate's `eps_F = 2|F| / (|J×B| + |∇p| + floor)` is bounded above by 2 by construction and sits at 2 wherever the denominator collapses — in vacuum, wherever any current remains. On the bundled `input.solovev` (pressure peaking at 0.125 Pa) it reports L2 1.969 and L∞ 2.000, the ceiling, while the same state's vacuum-safe normalized force error is 1.22e-2 and its dimensional ⟨|F|⟩ is 40 N m⁻³. Every certificate now also reports the quantities that cannot saturate, over `s ∈ [0.1, 0.99]` and the whole domain: ⟨|F|⟩/⟨|∇p|⟩ (Panici et al. 2023; `n/a` in vacuum), the vacuum-safe |F|/⟨|∇(B²/2μ₀)|⟩ that DESC's `ForceBalance` uses, ⟨|F|⟩ in N m⁻³, and the near-axis/bulk/edge split. The CLI prints them beside `eps_F`; `PolishReport` carries them; the artifacts record them.

**31.2-R3 / R6.** The README's "26-fold" (5.05e-2 → 1.91e-3) mixed the polish gain with an export-mesh and lift difference. The like-for-like measurement on the lifted native state is recorded with deck hash and commit: 7.12× in the bounded metric, 1.32× whole-domain and 1.06× on `s ∈ [0.1, 0.99]` in ⟨|F|⟩/⟨|∇p|⟩. The figure caption now states that its two cases were selected because polishing improves them and that this is not evidence of a general advantage.

Also: a committed certificate baseline with a regeneration tool (`tools/make_strong_force_baseline.py`), the output destination redacted from the strong-polish record, and the high-order-force-balance page derives the bound.

**Merge order.** Touches `vmex/core/strong_force.py`, which #258 (angular tail and nestedness margin) and #261 (batched sweep) also change; rebase after those land (plan section 6, step 16).


rogeriojorge:
Superseded by a fresh PR from the same branch: GitHub stopped creating workflow runs for this PR after its rebase.

PR #307, opened 2026-09-13
Reuse solver executables across same-device optimization trials

rogeriojorge:
## Problem and change

Perturbation predictors and axis retries can produce arrays with different JAX commitment flags even when their shapes, dtypes, and device placement are identical. The CLI solver then compiles extra `_block_lane` executables during optimization trials. Normalize the existing single-device carry/runtime arguments once after the donation-safe carry copy. An all-leaf guard preserves mixed-device and named-sharding layouts; whole-loop JIT, physics, tolerances, iteration limits, and every retry remain unchanged.

This is a focused, stacked follow-up to [PR #299](https://github.com/uwplasma/vmex/pull/299), targeting `perf/reuse-objective-linearization`. It adds seven production lines in one existing file. Two existing test files verify real executable reuse, input-buffer preservation, mixed-layout preservation, and successful consumption of prefetched executables.

## Reproduction and source context

Base: `de93e5e2`; candidate: `8a29ec0f53b630fa7866f0d6de363f27862a748c`.

The immutable [review evidence, `rejected_trial_replay` entry](https://github.com/uwplasma/vmex/blob/47de5a8baead9b3701e3eb890b844463eae95a4a/benchmarks/review_optimization_20260912.json) contains the exact public input hash, eight-parameter trial vectors, call order, settings, expected norms, and results. PR #299 contains the runnable `vmex.optimize.make_problem` recipe. An independent replay verified all seven residual/Jacobian norms, with maximum relative difference 2.66e-9.

The reduced QP problem uses `examples/data/input.minimal_seed_nfp2`, nfp=2, ns=13, mpol=ntor=3, a 12×8 angular grid, max_mode=1, implicit block Jacobians, perturbation warm starts, refine_tol=inf, and CLI mode. Its terms are the quasisymmetry-ratio residual on s=[0.4,0.9] with helicity (0,1), plus aspect ratio targeting 10 with residual weight 0.01. The recorded sequence recomputes accepted-point derivatives so that the predictor policy is retained; it replaces the outer optimization search with saved trial vectors.

## Evidence and qualified performance result

On a Xeon W-2295 CPU with JAX 0.9.2, actual `_block_lane` cache misses fell **11→3**, retaining the three necessary physics structures. Trace inspection distinguished commitment flags from treedef, shape, dtype, weak type, sharding, and function identity. The large rejected trial still executes all 12 solver-loop calls across predictor, hot-start, cold-start, and axis recovery; each fails at iteration 1 with the same bad-Jacobian status.

| Recorded component | Control | Initial candidate |
|---|---:|---:|
| Rejected trial residual | 18.930 s | 14.009 s |
| Its backend compilation | 12.866 s | 8.447 s |
| Next accepted residual | 4.856 s | 0.353 s |
| Following accepted residual | 4.001 s | 0.345 s |

These are instrumented component measurements, not a universal cold-start, GPU, or full-optimization speedup claim. The initial derivative varied from 43.039 to 52.143 s; total process envelopes were approximately 86 and 84 s. The timed candidate used a root-only single-device guard; the final guard additionally checks every leaf and preserves mixed layouts. For the measured single-device case, it executes the same normalization. Peak measured RSS did not increase (2.45→2.34 GB), but this is not a general memory bound.

Control/candidate replay produced bit-identical four residual vectors, three Jacobians, six full force histories, and 36 state-family arrays, with identical iterations, status, and recovery decisions. Final-source GPU QA/QI cold, small-boundary-step, return, and invalid-predictor recovery probes preserved status/iterations and force certificates: QA maximum absolute state difference 1.36e-20; QI maximum absolute difference 2.84e-10, with lambda relative difference below 9.84e-10. GPU timings are excluded because the device was shared. These small-step probes do not substitute for the large QP replay.

## Validation and review

- Local CPU: five runtime-key tests passed.
- Two-device CPU: seven tests passed, including real lane cache reuse `[1,1]`, donation/input retention, mixed NamedSharding preservation, successful prefetched execution, and explicit second-device placement.
- GPU: five runtime-key tests and explicit forward CPU/GPU parity passed; GPU-default policy and multigrid device-threshold tests passed in a separate correctly unpinned run.
- Changed executable production-line coverage: **3/3 (100%)**; both normalization and mixed-layout skip behavior exercised.
- Ruff, mypy, and `git diff --check` passed. Independent source/test review found no blocker.

Test-harness corrections were resolved without changing tolerances: the CPU fixture needed JIT enabled, and the device-policy test needed platform environment pins removed. CI remains pending for this draft. No merge or auto-merge is requested.

## Rationale and limits

[JAX array commitment](https://docs.jax.dev/en/latest/_autosummary/jax.Array.committed.html) and [explicit `device_put` placement](https://docs.jax.dev/en/latest/_autosummary/jax.device_put.html) document the placement semantics used here; the compile-cache attribution comes from the recorded traces. Normalization follows the existing alias-breaking copy and uses the selected sharding. It does not introduce a global device policy, bypass failed-equilibrium recovery, or change mathematical residuals.

Review the seven-line placement change and targeted regressions independently of PR #299's broader optimization work. The remaining expensive initial Jacobian and other optimization bottlenecks remain tracked there.


rogeriojorge:
Superseded by #319, which re-landed these seven lines on main with the reorder that removes the extra compile this PR caused.

PR #335, opened 2026-09-14
Factor once for reverse Jacobian rows and pull the gradient back once

rogeriojorge:
Follows #330 (B3a, the exact block adjoint), merged as `b5f84b10`. This PR targets `main`, with main merged in at `670419ea` without rebasing.
- **Conflicts:** only in `CHANGELOG.md` (both entries kept) and `benchmarks/INDEX.md` (regenerated; `--check` passes).
- **`vmex/core/implicit.py`:** #330 was squash-merged, so git reported conflicts. I took main's file and reapplied this PR's own diff (`9664cfa0..409be538`), which applied cleanly.
- **Result:** the diff against main is exactly this PR's own five files.

Plan brief B3b: the reverse Jacobian lane after #328. This covers `minimize(objective_terms)` and `implicit_jacobian_method="reverse_adjoint"`.

## User path

`minimize(objective_terms, ...)` builds its problem with `jac_solver="reverse"`. SciPy calls `problem.value_and_grad` each iteration.
- **Before:** that call assembled the full reverse Jacobian and formed `J^T r`. The Jacobian was #328's chunked map of the implicit rule's pullback, which re-factors the raw block Jacobian in every chunk.
- **After:** the gradient pulls back the residual once, through one factorization.
- **Reverse Jacobian:** `problem.jax_residual_jac` and the host `residual_jac` factor once per point, then pull the rows back in tangent-lane batches.

## Why the factorization has to move

The hoisting probe on office (untimed) mapped the implicit rule's pullback over chunks. Warm times were:

| map iterations | warm seconds |
|---:|---:|
| 1 | 3.24 |
| 8 | 14.77 |
| 64 | 114.94 |

That is 1.77 s per extra iteration, against 1.52 s for one raw block factorization. XLA does not hoist the linearization and factorization out of the map, so each chunk pays for both. On the seed QA problem (6722 residual rows, 8 dofs) the default width of 8 means 841 factorizations per reverse Jacobian.

## Change

- **`implicit._block_state_pullback`:**
  - Linearizes the raw residual and factors the block Jacobian once, then returns a per-row pullback.
  - Each row gets the exact block adjoint refined once, NaN where the certificate misses, then `-mu^T dF_raw/dp` plus the direct boundary term, the same as the scalar rule.
  - `implicit_state_pullback_multi_rhs(solver="block")` now uses it. Its parameter pullbacks are chunked with the solves instead of vmapped over the whole batch, and rows cross the map as flat vectors because `chunk_map` cannot reshape the zero-size families of a symmetric state.
  - The GCROT branch is unchanged.
- **`reverse_pullback` in `_least_squares_implicit`:**
  - Solves once and pulls each residual row back to the state and the direct `x` path.
  - Sends the state cotangent through the shared factors and maps the parameter cotangent back through `params_of`, in batches of `ndof` or `jacobian_batch_size`.
  - A failed trial keeps the penalty rows' direct derivative and a zero implicit pullback, as `solve_implicit_status` does.
- **Why not one call to the public helper:** calling `implicit_state_pullback_multi_rhs` with all rows would materialize the `rows x state` cotangent batch. On seed QA that is 6722 x 9300 doubles per buffer, which would undo #328's memory bound, so the reverse lane maps the helper's per-row pullback instead.
- **Single-VJP gradients:**
  - The traced `jax_value_and_grad` (`residual_value_and_gradient`) on the reverse lane, and on the block lane when its Jacobian misses the certificate, pulls back `r` once. This is the follow-up #328 named.
  - The host `value_and_grad` does the same when `jac_solver="reverse"` is set explicitly, as `minimize(objective_terms)` and `implicit_jacobian_method="reverse_adjoint"` do. A non-finite result falls through to the Jacobian lane, which keeps its retries and typed errors.
  - An automatic one-row problem keeps its host Jacobian: one row already costs one pullback, and the least-squares memo (`last_jac_key`) that `test_certified_trial_guards_reject_stale_or_missing_memo` checks stays populated.
  - The host block lane's fallback still goes through `jac_fn`, because the least-squares memo and the warm-start stash share it.
- **Counters:** a compiled shared pullback marks the adjoint counters unknown, like the compiled scalar rule, through the new `_mark_compiled_adjoint`.
- **Two factorizations per new point since #338:** a gradient at a new point now builds two raw block factorizations.
  - B1c's refinement factors at the descent's start iterate, as the preconditioner for its Newton steps.
  - This PR's reverse lane factors at the refined anchor, for the exact adjoint.
  - They cannot be shared, because they linearize different states.
  - `test_reverse_lane_factors_once_per_point` solves and refines at x first, then counts only the reverse lane's factorization.

Files: `CHANGELOG.md` +4/−0; `tests/manifest.json` +1/−0; `tests/test_optimize.py` +73/−0; `vmex/core/implicit.py` +104/−49; `vmex/core/optimize.py` +91/−15.

## Evidence

Office rows (Xeon W-2295, 36 threads, JAX 0.11.1, 4 threads, compilation cache disabled, fresh process per row, `~/vmex-agents/.heavy.lock`, 1-minute load ≤ 24 at start and end). Problem: #328's trace problem, `input.minimal_seed_nfp2` at mpol = ntor = 5 with `benchmarks/optimization.py`'s QA terms, max_mode = 1, ESS, `implicit_jacobian_method="reverse_adjoint"`. Base is #330's head `9664cfa0`, B3b is `409be538`. Raw block factorizations are counted at run time with a debug callback on `_raw_block_system`.

The rows are running on office under the heavy lock. The base row recomputes the full reverse Jacobian, about 841 factorizations, so it takes roughly 25 minutes. The table follows in a comment and replaces this paragraph.

Adjoint Krylov iterations are 0 on both trees: both use the exact block adjoint from #330. The adjoint counters report `None` for these compiled programs, so the factorization count is the measured cost driver.

## Single-stage per-trial cost is now mostly refinement

On #330's fixed 10-trial single-stage row, main took 12 trials, 23,754 refinement GCROT iterations in total, against 219 descent iterations per solve. B3a took 23,552. With the adjoint no longer dominant, refinement is most of each trial's cost. Those rows were measured before #320's early stop, and refinement is B1c's next target. B3b does not change the single-stage example's derivative path.

**Since then:** #338 (B1c) replaced that refinement. On the same 10 fixed trials, refinement went from 23,754 Krylov iterations to 80 GMRES iterations, refinement seconds fell to 0.34×, and per-trial time to 0.72×.

## Tests

- **`tests/test_optimize.py::test_reverse_lane_factors_once_per_point`** (new, `pr-implicit-response-b`):
  - `problem.jax_residual_jac` runs one raw block factorization at run time with a batch width of 1. The old map ran one per row.
  - That Jacobian equals the row-by-row pullback of `problem.jax_residual` to 1e-9.
  - The host and traced `jax_value_and_grad` gradients match `J^T r` to 1e-9, with one factorization.
  - The block lane's certificate fallback gives the same gradient.
- **Also run:** `test_least_squares_implicit_jac_solver_block` (reverse against per-dof GMRES, 1e-6), `test_uncertified_jacobian_falls_back_or_raises_with_evidence`, `test_certified_trial_guards_reject_stale_or_missing_memo`, and all of `tests/test_implicit_multi_rhs.py`, including #330's dense raw-adjoint check of the block pullback.
- **Results:** on JAX 0.11.1 (laptop), all pass except `test_certified_trial_guards_reject_stale_or_missing_memo`. It failed on the first host shortcut and passes after narrowing it to explicit `jac_solver="reverse"`. It passes on clean main `c4329b75` too. JAX 0.9.2 runs on office after the rows; the result follows in a comment.
- **Preflight:** `tools/preflight.py --static` passes ruff, mypy, docs prose and 79 of the 80 guard tests. The remaining guard, manifest ownership, fails only because `tests/test_tracing.py` collects nothing without `essos`, which the laptop venv lacks. It fails identically on #330's tree, and this PR's new selector collects.


rogeriojorge:
**Seed-QA rows: #330's head `9664cfa0` against B3b `409be538`.**

Setup:
- office Xeon W-2295, JAX 0.11.1, 4 threads, compilation cache disabled, `~/vmex-agents/.heavy.lock`, fresh process per row;
- #328's trace problem: `input.minimal_seed_nfp2` at mpol = ntor = 5 with `benchmarks/optimization.py`'s QA terms, max_mode = 1, ESS, `implicit_jacobian_method="reverse_adjoint"`;
- 6,722 residual rows, 8 dofs;
- raw block factorizations are counted at run time with a debug callback on `_raw_block_system`.

| row | #330 `9664cfa0` | B3b `409be538` |
|---|---|---|
| `problem.value_and_grad`, first call (compile included) | stopped after 4,111 s, not finished | 48.4 s, 1 factorization |
| `problem.value_and_grad`, second point | not run | 34.3 s, 1 factorization |
| `problem.jax_residual_jac`, 6,722 × 8, first call | not run | 238.9 s, 1 factorization |
| peak RSS | not recorded | 1.46 GiB (gradient row), 1.52 GiB (Jacobian row) |
| load start → end | 8.32 → 8.33 when stopped | 8.57 → 9.91 (gradient), 9.91 → 8.32 (Jacobian) |

- **What #330's base row computes:** on #330, `value_and_grad` on the reverse lane assembles the full reverse Jacobian. That is #328's chunked map of the implicit rule's pullback, with one raw block factorization per 8-row chunk, 841 per call. The row was stopped by decision after 4,111 s (SIGTERM), not finished.
- **Lower bounds:** B3b's reverse Jacobian is at least 17× faster (238.9 s including compile, against more than 4,111 s). B3b's first gradient call is at least 85× faster (48.4 s against more than 4,111 s).
- **Adjoint Krylov iterations:** 0 on both trees, because both use #330's exact block adjoint. The adjoint counters report `None` for these compiled programs. The factorization count is the measured cost driver: 1 per call on B3b, 841 per call on #330.
- **Where B3b's gradient call spends its time:** mostly today's Krylov refinement, 4,000 GCROT iterations per call on this deck. #338 replaces that refinement.

**Consistency.**
- **B3b against itself:** B3b's two rows agree. The value is 0.6268904871584033 in both, and the gradient norm is 5.590748889048 (host `value_and_grad`) against 5.590748889050 (`J^T r` from the reverse Jacobian).
- **Against #330:** one untimed `problem.value_and_grad` on #330's tree through the default block lane (`J^T r` from the block Jacobian, 63.4 s, 1 factorization) gives the identical value, 0.6268904871584033. Its gradient agrees with B3b's reverse lane to 1.8e-13 relative (largest entry difference 8.6e-13).

**JAX 0.9.2 focused tests (office, `~/venvs/vmex-ci-jax092`, `409be538`): 12 passed, 4 skipped (the full-marked multi-RHS tests) in 737 s.**

Selection:
- `tests/test_optimize.py`: `test_reverse_lane_factors_once_per_point`, `test_least_squares_implicit_jac_solver_block`, `test_uncertified_jacobian_falls_back_or_raises_with_evidence`, `test_certified_trial_guards_reject_stale_or_missing_memo`;
- all of `tests/test_implicit_multi_rhs.py`.

The same selection passed on JAX 0.11.1 on the laptop.

PR #360, opened 2026-09-15
Move a rung's residual continuation to the device its state goes to

rogeriojorge:
## Failing command

On the office workstation (2x RTX A4000, JAX 0.11.1 + CUDA 13), running the GPU
workflow's own selection for the first time since 2026-07-31:

```console
$ pytest -q $(python tools/test_manifest.py select gpu-smoke)
FAILED tests/test_gpu_ci.py::test_multigrid_auto_moves_state_across_policy_threshold
1 failed, 65 passed, 6 skipped in 2168.24s (0:36:08)

E  ValueError: Received incompatible devices for jitted computation.
E  Got argument carry.state.R_cos of _while_lane with shape float64[11,6]
E  and with device ids [0] on platform GPU and argument carry.fsqr of
E  _while_lane with shape float64[] and with device ids [0] on platform CPU
vmex/core/solver.py:2331: ValueError
```

## This is a regression inside the 0.9.0 window

```console
v0.8.1  ->  1 passed in 11.57s
main    ->  1 failed in 12.10s
```

`git bisect run` over `v0.8.1..main` on the GPU box names
**b208df4f (#321, B4b)** as the first bad commit.

## Cause

#321 wrapped `_constraint_baselines` so it commits its arguments and clears the
default-device context, keeping commitment-sensitive lanes to one executable per
structure. That is correct on its own terms, and it removed an accident: the
ambient `device_context` used to drag the whole carry onto one device.

What it exposed is in `solve_multigrid`. Each rung places its state:

```python
state = _put_numeric_leaves(state, _placement_device(device, resolution))
```

but the previous rung's residual scalars, captured at `multigrid.py:570` as
`residual_continuation = (carry.fsqr, carry.fsqz, carry.fsql)`, were passed
straight into the next rung's `_initial_carry`. With `device="auto"` a coarse
rung below `GPU_MIN_ITERATION_WORK` runs on CPU and the fine rung above it on
GPU, so the carry arrives with its state on GPU and its residual scalars on CPU.

`commit_to_single_device` cannot fix this downstream: by contract it returns the
tree unchanged unless every array leaf already shares one `SingleDeviceSharding`.
It unifies commitment, it does not move arrays between devices.

The free-boundary driver already does the right thing for its own continuations
(`multigrid.py:832`, `constraint_continuation = _put_numeric_leaves(...)`). This
gives the fixed-boundary driver the same treatment.

## Why nothing caught it

`tests/test_gpu_ci.py::test_multigrid_auto_moves_state_across_policy_threshold`
is correct and would have caught this the day it landed. It runs only in
`gpu.yml`, which is `workflow_dispatch`-only against a self-hosted runner --
and `gh api repos/uwplasma/vmex/actions/runners` returns an empty list. The last
successful GPU run was 2026-07-31, six weeks before #321.

So this PR also adds a guard that needs no GPU.
`test_multigrid_moves_residual_continuation_with_the_state` forces consecutive
rungs onto two devices, which reproduces the same mixed carry on two CPU
devices. It joins `pr-device`, the lane already running under
`XLA_FLAGS=--xla_force_host_platform_device_count=2`:

| tree | result |
|---|---|
| this branch | `1 passed in 74.23s` |
| `vmex/core/multigrid.py` reverted | `1 failed in 45.47s` -- `ValueError` |

## Verification

- The originally failing GPU test on the office box with this fix applied:
  `1 passed in 9.70s`.
- The full `gpu-smoke` selection re-run on the office box (result in the first
  comment when it lands; the pre-fix baseline was 1 failed, 65 passed, 6 skipped).
- CPU side, laptop: `tests/test_device.py test_gpu_ci.py test_implicit_device.py
  test_multigrid_ladder.py test_runtime_recompile_keys.py test_trace_budgets.py`
  -> 69 passed, 31 skipped, 2 xfailed.
- `python tools/preflight.py --static` -> PASS (ruff, mypy on 76 files, docs
  prose, 85 guard tests).

## What a reviewer should check

- That moving the continuation does not change the numbers on a same-device
  ladder: `_put_numeric_leaves` is a no-op when the leaf is already on the
  target (`if sources == {device}: return leaf`), so a CPU-only or GPU-only run
  is untouched. The trace-budget and recompile-key tests above cover the
  executable count.
- Whether the same gap exists for any other value carried between rungs. I
  checked `constraint_continuation` (already moved on the free-boundary path,
  and `None` across a grid change on the fixed path) and the state itself.
- The real fix for the class is a GPU runner. Until `gpu.yml` has one, every
  `tests/test_gpu_ci.py` assertion is unexecuted, and #321 sat broken for a day
  only because this release happened to re-run it by hand.


rogeriojorge:
**Full `gpu-smoke` lane on the office workstation with this fix applied** (2x RTX A4000, JAX 0.11.1 + CUDA 13, `JAX_ENABLE_X64=1`):

```
66 passed, 6 skipped in 2116.88s (0:35:16)
```

against the pre-fix baseline on the same machine and environment:

```
1 failed, 65 passed, 6 skipped in 2168.24s (0:36:08)
FAILED tests/test_gpu_ci.py::test_multigrid_auto_moves_state_across_policy_threshold
```

No other test changed state, and the wall time is flat (35:16 vs 36:08), so the
extra `_put_numeric_leaves` call costs nothing measurable.

This is the first fully green GPU run since 2026-07-31.

Note GPU 0 was carrying an unrelated job at 14.9 GB of its 16.4 GB throughout,
so this run shared the box; `nvidia-smi` showed this process at 468-582 MiB on
GPU 0 and 178-520 MiB on GPU 1. The two-GPU assertions still exercised both
devices.

PR #366, opened 2026-09-18
Accelerate winding objective sensitivities

dseidita1919:
# What this changes

The winding optimizer previously stopped after four LBFGSB iterations but differentiated the result as a converged root. Its implicit response also used JAXopt's default normal-equation CG, and the outer VMEX Jacobian was forced onto the block-forward path because that response was not transposable.

This PR implements the first resolution-independent performance slice:

- raises the inner LBFGSB cap from 4 to 100 and certifies convergence with LBFGSB's infinity-norm optimality criterion;
- replaces normal-equation CG with a certified GCROT solve of the original optimality-Jacobian system, wrapped in `jax.lax.custom_linear_solve` with a transpose solve;
- restores `implicit_jacobian_method="auto"`, so the current scalar objective uses a nested reverse adjoint;
- exploits exact field-period block-circulant symmetry by assembling one target-period block row, applying a block DFT, and taking independent sector SVDs;
- retains every singular value and falls back to the full matrix when the toroidal grid is not divisible by `nfp`;
- keeps plasma geometry in the Fourier-coefficient domain so the symmetry reduction is exact for derivatives, not only values.

The symmetry path is parameterized by `nfp`, `WINDING_NPHI`, and `WINDING_NTHETA`; it does not assume the current 16×16 grid.

# Evidence

## Correctness

The new tests verify:

- full-spectrum agreement for `nfp = 1, 2, 3, 4` and the non-divisible-grid fallback;
- equality of the full and sector formulations through value, gradient, and HVP for the current `nfp=2` case;
- forward and transpose solves for symmetric and nonsymmetric systems;
- JVP, reverse gradient, and HVP agreement for an implicitly solved root.

A representative 25-variable winding solve required 38 LBFGSB iterations to reach `8.06e-7` projected optimality, confirming that `maxiter=4` was not a safe root assumption.

## Warm CPU timings

JAX 0.11.2, float64, current 16×16 winding/plasma grid:

| Kernel | Full matrix | Period sectors | Speedup |
|---|---:|---:|---:|
| Spectral value | 10.12 ms | 6.15 ms | 1.65× |
| Spectral gradient | 35.88 ms | 21.67 ms | 1.66× |
| Spectral HVP | 43.27 ms | 19.42 ms | 2.23× |

For the representative 25-DOF response, default normal-equation CG took 503.3 ms versus 262.3 ms for direct-system GCROT (1.92×).

## Commands run

```console
ruff check examples/optimization/QA_winding_optimization.py tests/test_winding_optimization.py
# All checks passed

pytest -q tests/test_winding_optimization.py
# 9 passed

pytest -q tests/test_winding_optimization.py tests/test_examples.py tests/test_optimize.py
# 85 passed, 38 skipped

VMEX_EXAMPLES_CI=1 JAX_ENABLE_X64=1 python examples/optimization/QA_winding_optimization.py
# completed successfully through optimization, final solve, and output generation
```

# Follow-up scope

This PR intentionally leaves the specialized scalar entropy JVP and an explicit small optimality-Jacobian factorization for later profiling. It establishes the exact symmetry-reduced and adjoint-capable architecture first.

# Checklist

- [ ] `python tools/preflight.py` passes locally (targeted relevant suites were run above).
- [ ] Changed executable lines are at least 95% covered by the repository diff-coverage command.
- [x] No VMEC physics-schema change.
- [x] Performance claims are backed by the timings and correctness tests above.

# Multi-case benchmark matrix (2026-09-18)

The committed [benchmark report](https://github.com/uwplasma/vmex/blob/ds/winding-adjoint-performance/benchmarks/winding_surface_optimization.md) contains the full protocol and raw-data link. It covers three fixed geometries at 8×8, 12×12, 16×16, and 24×24, plus two nested-optimization geometries at 12×12 and 16×16. Every plot shows medians and the complete case range; all samples and environment metadata are retained in JSON.

Runtime scaling

Fidelity

Resolution-dependent speedup

- Standalone geometric-mean speedups grow from 1.20–1.55× at 8×8 to 2.54–2.90× at 24×24, so the performance gain is explicitly resolution-dependent.
- Across all standalone cases/resolutions, maximum relative errors are 7.09e-15 for gradients and 6.36e-13 for HVPs.
- All current maxiter=100 nested solves certify stationarity below 1e-6. Against a full-SVD, dense-25×25 implicit reference, the worst relative outer-gradient error is 7.58e-8 across entropy-only, inverse-distance-only, and mixed objectives.
- The historical four-iteration solves fail stationarity (3.02e-2 to 1.00), so their lower primal time is not treated as an accuracy-matched speed result.
- The nested mixed-gradient path is 11.75–35.54× faster than the historical forward path; at matched maxiter=100 fidelity, the periodic primal solve is 1.65–2.23× faster than the full-SVD reference.


rogeriojorge:
The c2 parity lane fails on `tests/test_test_manifest.py::test_collected_suite_has_exact_manifest_ownership`: the new `tests/test_winding_optimization.py` has no entry in `tests/manifest.json`, so its nodes are collected without lane ownership. Adding one row there (category/kind/cost/device/assets/reference plus the lanes it should run in) is all the lane needs; keep the PR-lane cost in mind, since the fast lane has an 8-minute cap and the parity lanes 25.

The rest of the run is green.

dseidita1919:
## PR #367 comparison, compatibility, and combined benchmark

I independently benchmarked PR #367 (`0f4ba71e`) against the historical implementation (`3f53e074`), PR #366 before this update (`9ab36917`), two dense full-SVD references, and the selected combined implementation. The complete protocol, every retained sample, provenance, and all figures are in the [benchmark report](https://github.com/uwplasma/vmex/blob/681be53824126946e05814c079756050aee20e1e/benchmarks/winding_surface_optimization.md), [primary raw record](https://github.com/uwplasma/vmex/blob/681be53824126946e05814c079756050aee20e1e/benchmarks/winding_surface_pr367_comparison_20260922.json), and [candidate-screen record](https://github.com/uwplasma/vmex/blob/681be53824126946e05814c079756050aee20e1e/benchmarks/winding_surface_pr367_candidates_20260922.json).

### Are the methodologies the same?

They overlap, but they are not equivalent:

| concern | PR #366 | PR #367 |
| --- | --- | --- |
| inner solve | converged LBFGSB, `maxiter=100` | converged LBFGSB, `maxiter=200` |
| implicit derivative | custom JVP with certified, transposable matrix-free GCROT | native JAXopt VJP with dense 25×25 LU |
| outer differentiation | `auto` (reverse for the current scalar residual) | forced `reverse_adjoint` |
| normalization scales | derivative stopped | derivative retained |
| induction SVD | exact field-period block decomposition | original full-torus SVD |
| pairwise geometry | exact one-source-period reduction | original full pair tensors |

Both eliminate the historical 25-way forward outer response, so most of PR #367's speedup over the original is the same high-level reverse-mode improvement already present in #366. PR #367's live normalization-scale derivative is a distinct fidelity correction. PR #366's periodic SVD and pairwise reductions are distinct, resolution-general performance improvements. Dense LU and custom GCROT are alternative implicit solvers rather than additive optimizations.

### Benchmark protocol

- Standalone: nominal, close-clearance, and shaped geometries at 8×8, 12×12, 16×16, and 24×24; nine retained warm samples for value, gradient, and HVP.
- Nested: nominal and shaped geometries at 12×12 and 16×16; five retained warm samples for entropy-only, inverse-distance-only, and equally mixed outer gradients.
- Fidelity: original full SVD/full pair sets for standalone work; independent dense 25×25 implicit solves with both frozen and live normalization scales for nested work.
- Independent derivative check: two directions and five-point finite differences at `1e-3` and `3e-4` steps.
- JAX 0.11.2, float64, CPU, disabled compilation cache; no case- or resolution-specific tolerance tuning. Every recorded timing and numeric result is finite.

### End-to-end speed

| case | resolution | historical, unconverged (ms) | live full-SVD dense ref. (ms) | PR #366 (ms) | PR #367 (ms) | combined (ms) | #367 / combined |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| nominal | 12×12 | 3,868.23 | 493.50 | 228.83 | 372.28 | 194.18 | 1.92× |
| nominal | 16×16 | 15,581.56 | 1,979.11 | 502.19 | 1,193.04 | 470.36 | 2.54× |
| shaped | 12×12 | 3,102.47 | 543.80 | 252.17 | 417.79 | 237.52 | 1.76× |
| shaped | 16×16 | 17,657.79 | 2,324.22 | 629.00 | 1,536.30 | 678.49 | 2.26× |

Geometric-mean speedups versus the accuracy-matched live full-SVD reference are **2.87× for #366, 1.44× for #367, and 3.03× for the combination**. The combination is **2.10× faster than #367**. Relative to #366 it is 1.055× faster geometrically, but I treat that as preserved performance rather than a new speed claim: it is faster in three cases and 7.9% slower in shaped 16×16.

The historical path is 21.76× slower than the combination, but its stationarity residuals are `3.02e-2`–`1.00`; that is not an accuracy-matched comparison.

Runtime scaling

### Fidelity

All #366, #367, combined, and dense-reference solves satisfy the `1e-6` stationarity criterion (`6.81e-7`–`9.04e-7`). Worst errors against the live-scale dense reference are:

| implementation | max observable absolute error | entropy gradient rel. L2 | inverse-distance gradient rel. L2 | mixed gradient rel. L2 |
| --- | ---: | ---: | ---: | ---: |
| PR #366 | 7.18e-11 | 7.12e-3 | 1.27e-1 | 5.61e-2 |
| PR #367 | 0 | 3.06e-13 | 2.16e-12 | 2.61e-12 |
| combined | 7.18e-11 | 1.05e-8 | 7.75e-8 | 5.97e-8 |

The frozen-scale dense reference reproduces #366's discrepancy, so this is specifically the omitted normalization-scale derivative—not the periodic reduction or GCROT solve. Retaining live scales removes that bias in the combined path.

Standalone combined fidelity remains at roundoff: spectral maxima are `8.33e-17` absolute value error, `7.09e-15` relative gradient error, and `6.36e-13` relative HVP error; pairwise maxima are `8.33e-17`, `6.78e-15`, and `1.73e-13`, respectively.

At the `1e-3` finite-difference step, maximum relative directional errors are:

| implementation | entropy | inverse distance | mixed |
| --- | ---: | ---: | ---: |
| PR #366 | 1.07e-1 | 7.16e-2 | 4.68e-2 |
| PR #367 | 2.44e-4 | 8.80e-4 | 6.00e-4 |
| combined | 5.89e-4 | 8.77e-4 | 5.98e-4 |

Thus both live-scale methods agree with finite differences to below 0.1% for entropy-only, inverse-distance-only, and mixed objectives.

Nested and standalone fidelity

Five-point directional fidelity

### Compatibility screen

The source edits conflict textually because both PRs modify the same inner-solve function, but the algorithms are compatible. I measured three combinations:

1. `live_gcrot`: #366 plus live normalization scales.
2. `native_lu`: the direct/trivial combination of #366's periodic kernels with #367's native VJP and dense LU.
3. `native_gcrot`: periodic kernels plus a native JAXopt VJP supplied with certified GCROT.

| variant | nominal 12 (ms) | nominal 16 (ms) | shaped 12 (ms) | shaped 16 (ms) | geometric mean (ms) | #367 / variant | max mixed-gradient rel. error |
| --- | ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| PR #366 | 203.98 | 431.81 | 292.36 | 673.91 | 362.95 | 2.01× | 5.61e-2 |
| PR #367 | 340.37 | 1,052.70 | 525.54 | 1,494.64 | 728.36 | 1.00× | 2.61e-12 |
| live GCROT | 198.14 | 426.91 | 250.45 | 664.47 | **344.45** | **2.11×** | 5.97e-8 |
| native LU | 204.21 | 452.44 | 226.85 | 710.12 | 349.28 | 2.09× | 1.56e-8 |
| native GCROT | 216.75 | 418.16 | 248.85 | 647.95 | 347.69 | 2.09× | 5.97e-8 |

The trivial native-LU combination preserves both branches' essential benefits, but is 1.4% slower geometrically than `live_gcrot` and replaces a working matrix-free path with dense factorization. Native GCROT also gives no aggregate advantage over the existing custom rule. Since these differences are near timing scatter, the selected implementation is the smallest change: keep #366's `maxiter=100`, exact periodic kernels, and certified custom GCROT JVP, while removing only the normalization-scale `stop_gradient` identified by #367.

Compatibility candidates

### Resolution scaling and scope

| resolution | spectral value | spectral gradient | spectral HVP | pairwise value | pairwise gradient | pairwise HVP |
| ---: | ---: | ---: | ---: | ---: | ---: | ---: |
| 8×8 | 1.10× | 1.51× | 1.16× | 1.28× | 1.05× | 1.25× |
| 12×12 | 1.56× | 1.67× | 2.01× | 1.60× | 1.83× | 1.53× |
| 16×16 | 1.75× | 2.17× | 2.32× | 1.38× | 1.18× | 1.30× |
| 24×24 | 2.34× | 2.61× | 2.85× | 1.55× | 1.79× | 1.95× |

The speedup magnitude is explicitly resolution-dependent, but every entry is above one. The implementation is not hard-coded to a resolution: it uses the exact periodic path whenever the toroidal grid divides by NFP and otherwise falls back to the full calculation.

The selected change is also objective-agnostic. It corrects the total inner derivative and accelerates the shared spectrum/pairwise work; it does not special-case the current `1/minimum_distance` outer residual.

### Published implementation and verification

- `681be53`: benchmark harness, raw records, report, and graphics.
- `3e5a9cc`: combined live-scale implementation.
- `5c733d0`: regression that prevents normalization scales from being silently removed from the total derivative.
- Local checks: **44 focused tests passed**; **76 broader tests passed, 38 expected skips**; Ruff, compile, benchmark-index/provenance checks, and the complete CI-mode QA winding example passed.

CI has been triggered on the updated PR branch. This update does not merge the PR.

PR #367, opened 2026-09-18
Converge the winding inner solve and take its derivative in reverse mode

rogeriojorge:
## Context

`examples/optimization/QA_winding_optimization.py` was slow to compile and its
Jacobian cost grew with the boundary dof count. Profiling it first showed a
larger problem: the derivative it handed scipy was not the derivative of the
residual scipy was evaluating.

Base commit `3f53e0740f5ec0a3a0e78149a188f1f16cd125d8` (`ds/working_winding`).
All measurements on macOS 26.6.2 / Apple Silicon (10 CPUs), JAX CPU backend,
float64, Python 3.13.7, jax+jaxlib 0.11.2, jaxopt 0.8.5, scipy 1.18.1,
numpy 2.5.3, essos 0.17, this checkout of vmex 0.9.1, with
`VMEX_COMPILATION_CACHE_DIR=disabled` and a fresh process for every number.

**Diagnosis.** The inner `LBFGSB` ran `maxiter=4` and stopped with a
projected-gradient norm of 4.8; the same solve converges at iteration 69 to
1.6e-6. `root_jvp` differentiates the optimality conditions of the *converged*
problem, so the Jacobian described a different map than the four-step residual.
Against central differences of `problem.residual` — whose noise floor is 6e-10,
measured independently with `opt.aspect_ratio` — the shipped Jacobian was wrong
by 16-157%, including sign flips, and the error was a *flat plateau* over
h = 1e-2 ... 1e-6, so it was not truncation error (figure 2, left panel).

Two separable causes, measured separately:

1. the four-step budget, dominant: changing only the inner linear solver
   (`root_jvp(..., solve=solve_lu)`) left the error at 17-122%;
2. `stop_gradient(scales)`: once the inner solve converges, restoring that
   sensitivity moves directional derivatives by 4-17%, and is what takes the
   remaining error from ~1e-1 to ~1e-3.

Separately, the hand-written `custom_jvp` could not be transposed: `jax.vjp`
raised `NotImplementedError` from
`jax/_src/lax/control_flow/solves.py::_linear_solve_transpose_rule`, because
jaxopt's default `solve_normal_cg` builds a `custom_linear_solve` with no
`transpose_solve`. That is why the example pinned `block_tridiagonal`.
`solve_lu` removes the obstruction, and also changes the answer: the iterative
default was 32% off the direct solve on the seed tangent.

## Changes

One file, +37 / -47 (a net deletion; most added lines are comments).

1. `_solve_winding_surface`: `maxiter=4` -> `maxiter=200`. This is a cap, not a
   budget: `tol=1e-6` is what stops the solve, at 32-97 iterations over the
   nine boundaries checked, never reaching the cap, with zero active bounds.
   Adds `implicit_diff_solve=solve_lu` (the optimality Jacobian is 25x25).
2. Deleted the `jax.custom_jvp` wrapper and `_solve_winding_surface_jvp`;
   jaxopt's own implicit rule supplies the derivative. Drops `import functools`
   and the `root_jvp` import.
3. `_winding_objective_value`: removed `jax.lax.stop_gradient(scales)`.
4. `implicit_jacobian_method`: `"block_tridiagonal"` -> `"reverse_adjoint"`,
   pinned rather than left to `"auto"` so uncommenting a dormant term
   (nres > 1, where `"auto"` switches to the forward block response) still
   reaches the reverse lane.

Derivatives are **implicit** — the exact adjoint of the converged inner optimum
— not finite differences and not an advertised approximation. The convergence
control did change, so the residual changes: 15.013 -> 10.801 at the seed,
because the winding surface is now actually optimized. Objective, weights,
constraints, quadrature, active modes, grids, the outer `least_squares` call
and every output file are untouched.

## Validation and performance

Harnesses, logs and raw JSON live outside the tree; every command below was run
from a fresh process with the persistent compilation cache disabled.

**Directional derivatives.** `J(x) @ (D v)` against `[r(x+hDv) - r(x-hDv)]/2h`,
`D = diag(PARAMETER_STEP * problem.scales)`, 3 boundaries x 5 directions
(3 random + 2 single coefficients), h swept 1e-2 ... 1e-6, with offsets,
scales, bounds and the inner solve recomputed on both sides.

| rule | inner solve | best-h relative error |
| --- | --- | --- |
| baseline `root_jvp(solve_normal_cg)` | `maxiter=4` | 1.6e-1 ... 1.6e+0 (flat in h) |
| `root_jvp(solve_lu)` only | `maxiter=4` | 1.7e-1 ... 1.2e+0 |
| patched, `stop_gradient(scales)` kept | converged | 2.1e-3 ... 1.9e-1 |
| **patched (as shipped)** | converged, `tol=1e-6` | **3.5e-5 ... 1.7e-1** |
| patched rule, tight reference | converged, `tol=1e-10` | **6.0e-8 ... 2.3e-5** |

The shipped row's large entries are all on the two single-coefficient
directions whose derivative is ~1e-4; their analytic values agree with the
tight reference to ~1e-5, so those entries are finite-difference noise, not
derivative error. The tight row is what certifies the rule (figure 2, right).

Cross-check: the reverse lane and the forward `root_jvp(solve_lu)` +
`block_tridiagonal` lane return the same Jacobian to ~1e-6 relative.
Determinism: `r(x_a), r(x_b), r(x_a)` agreed to the last digit everywhere.

**Inner convergence**, over the seed, both final geometries and six perturbed
boundaries: 32-97 iterations, all terminated by `tol` (error <= 9.7e-7,
projected gradient <= 2.3e-6), `maxiter=200` never reached, zero active bounds
in every case.

**Derivative cost** (medians over distinct nearby x; figure 3):

| max_mode | dofs | compile [s] | warm residual [s] | warm `residual_jac` [s] |
| 1 | 8 | 35.5 -> 25.2 (1.41x) | 0.44 -> 0.93 | 0.47 -> 0.89 (0.53x) |
| 3 | 48 | 44.4 -> 27.4 (1.62x) | 1.60 -> 1.73 | 7.31 -> 1.45 (5.0x) |
| 5 | 120 | 82.5 -> 30.3 (2.72x) | 2.96 -> 2.66 | 42.6 -> 3.76 (11.3x) |

Reference floor from the same harness with `opt.aspect_ratio` as the only term,
i.e. VMEX's own machinery: compile 26.8 / 27.5 / 36.3 s, Jacobian 0.19 / 1.23 /
4.40 s. After the patch the winding term adds **0.23 s** to the max_mode=3
Jacobian instead of 6.08 s (27x), and no measurable compile time instead of
16.9 s. The residual is slower at max_mode=1 because it now runs ~69 inner
iterations instead of 4; that is the price of the correct map, and it is
already repaid at 48 dofs.

**Whole script**, cold, two repetitions each:

| case | baseline | patched | ratio |
| --- | --- | --- | --- |
| max_mode=1, shipped settings | 74.45 / 64.83 s | 65.68 / 66.98 s | no measurable change |
| max_mode=5, scratch override | 294.1 s | 146.5 s | **2.01x** |

Peak RSS 2.86-3.82 -> 3.63-3.80 GiB (max_mode=1) and 3.87 -> 4.13 GiB
(max_mode=5). The larger-mode run also caught no shape or recompilation
regression; CPU time fell 1090 s -> 414 s.

**Optimization quality.** Both runs' final boundaries re-scored with the *same*
converged objective, plus a hard minimum plasma/winding distance on a 96x96
diagnostic grid (production stays 16x16). Equal budget, 10 function
evaluations, and the final `NS=51` equilibrium converged in both cases.

| | winding residual `1/d` | true min clearance [m] |
| --- | --- | --- |
| seed (max_mode=1 resolution) | 10.801 | 0.08908 |
| baseline final | 10.809 | 0.08899 |
| **patched final** | **9.842** | **0.09921** |
| seed (max_mode=5 resolution) | 10.738 | 0.08797 |
| baseline final | 9.701 | 0.08705 |
| **patched final** | **8.822** | **0.09247** |

At max_mode=1 the baseline does not move off the seed (10.801 -> 10.809) while
the patched run reaches 9.842. At max_mode=5 the patched run ends 9.1% lower in
residual and 6.2% wider in true clearance, in half the wall time. Every
clearance stays well above `MINIMUM_DISTANCE = 0.05`; no new guardrail
violations.

**Other checks.** `ruff check` clean; `python -m py_compile` clean; the
`VMEX_EXAMPLES_CI=1` smoke path exits 0 and writes all six figures and both
wout/indata artifacts; and a dormant-term smoke test (`opt.aspect_ratio` plus
the winding term, nres = 2) returns a finite 2x8 Jacobian through
`reverse_adjoint`, so uncommenting a legacy term still works.

## Figures

Generated locally from the recorded measurements; there is no authenticated
attachment path from this environment, so they are not embedded:

1. scipy cost against elapsed wall time from process start, including startup
   and compilation, at max_mode=1 and max_mode=5, annotated with the
   common-objective scores.
2. directional-derivative error against finite-difference step: baseline,
   patched as shipped, and patched against a noise-free tight reference.
3. warm `residual_jac` and cold compile time against boundary dof count, with
   the VMEX-only floor.

## Limits

- Not the 10x the task aimed for on the full script at the shipped
  `max_mode=1`, and it is not reachable by editing this file: of the 74.4 s
  baseline, ~32 s is work this patch cannot touch (imports and seed solve 3.9 s,
  problem construction 5.3 s, stage report 3.4 s, final `NS=51` equilibrium
  5.5 s, wout and six plots 14.4 s). A free derivative would still cap the
  whole-script gain at ~2.3x. The 10x is reached on the derivative itself where
  it hurts: 11.3x warm Jacobian at max_mode=5, 27x on the winding term's own
  Jacobian contribution at max_mode=3.
- CPU only (Apple Silicon). No GPU, no Linux, no multi-device run.
- The full nine-stage ladder was not run end to end; max_mode=5 was exercised
  as a single stage through a scratch override, with shipped settings untouched.
- `tests/test_examples.py::test_every_example_is_tested_or_listed_as_untested`
  already fails on `ds/working_winding` without this patch: the example is
  neither covered by a test nor listed in `UNTESTED_EXAMPLES`. Fixing that needs
  a `tests/` change, outside this one-file scope.
- With only the winding term active the boundary is unconstrained in QS and
  aspect ratio and moves a long way (max_mode=1: aspect 11.69 -> 4.68, QS total
  0.033 -> 1.72). That is the single active objective doing what it is told, not
  a quasi-axisymmetry result.
- The inner budget is a cap: a boundary needing more than 200 iterations would
  truncate silently and reintroduce the inconsistency. The most observed was 97.


rogeriojorge:
The misc parity lane fails on `tests/test_examples.py::test_every_example_is_tested_or_listed_as_untested`: `examples/optimization/QA_winding_optimization.py` is missing a test or an `UNTESTED_EXAMPLES` entry. The script already honours `VMEX_EXAMPLES_CI=1` (max_mode 1, 4 evaluations, ns=31), so a smoke test beside the other optimization-example tests is the right registration rather than an exemption.

Two things to clean before this reaches main:
- the shipped ladder is a debug budget — `MAX_MODES, MAX_NFEV = [1,], [10,]` with the real ladder commented out on the next line. A shipped example should run the budget it documents (same issue fixed in #350 for the LASYM QH example).
- the commented-out `XLA_FLAGS` and `jax_enable_compilation_cache` lines at the top: keep them only as a documented, active choice, otherwise drop them.

PR #369, opened 2026-09-19
Restore the QI sheet-current field builder the weekly gate imports

rogeriojorge:
## Failing job

[Weekly qi-gate](https://github.com/uwplasma/vmex/actions/runs/35425350853/job/105850215488) on `main` (weekly run 35425350853): **1 passed, 5 errors in 59 s** — every test that needs the `sheet_field` fixture errored at setup.

```
    @pytest.fixture(scope="module")
    def sheet_field(tmp_path_factory):
>       import build_qi_sheet_mgrid as builder
E       ModuleNotFoundError: No module named 'build_qi_sheet_mgrid'
```

## Cause

`tools/build_qi_sheet_mgrid.py` was removed in the 0.9.1 release merge (#363) as "referenced by nothing in the tree" — judged against two signals: no entry in `tools/README.md`, and no path reference anywhere. Both signals were blind here. `tests/test_qi_sheet_gate.py` is a *hard* consumer: its module-scoped `sheet_field` fixture calls `builder.build(outdir)` to construct the QI sheet-current mgrid in-session. But it reached the module by prepending `tools/` to `sys.path` and importing the bare name, so the only spelling of the dependency in the tree was an import name — invisible to a path-based review.

Nothing caught the removal because no pull-request or nightly lane collects these `full`-marked tests; the weekly `qi-gate` lane added in #355 is the first thing that ever runs them.

## Fix

- Restore `tools/build_qi_sheet_mgrid.py` byte-for-byte from the commit before its removal. This is the real fix: the gate's whole point is a self-consistent public QI free-boundary case compared against recorded VMEC2000 goldens across five radial rungs. Skipping instead would retire the gate, and the fixture is deterministic and needs no fetched asset — it builds from the bundled `examples/data/input.nfp2_QI`.
- Spell the dependency as a path: `importlib.util.spec_from_file_location(...)` against `ROOT / "tools" / "build_qi_sheet_mgrid.py"` instead of a `sys.path` search, so a future path-based orphan review sees it and a missing file fails with a message that names the file.
- Add the tool to `tools/README.md`, marked as a required fixture dependency rather than a convenience script.

No library code changes; no behaviour change to the gate itself.

## Local verification

macOS arm64, JAX 0.11.1, absolute `PYTHONPATH` to the worktree:

```
vmex .../tests/../vmex/__init__.py   (worktree, not the installed wheel)
jax 0.11.1

$ JAX_ENABLE_X64=1 RUN_FULL=1 pytest -q tests/test_qi_sheet_gate.py::test_sheet_field_confines
.                                                                        [100%]
============================= slowest 5 durations ==============================
123.78s setup    tests/test_qi_sheet_gate.py::test_sheet_field_confines
1 passed in 124.24s (0:02:04)
```

The fixture now builds (123.78 s), and `test_sheet_field_confines` passes its `fit_metric`, `alignment` and `phiedge` thresholds — so the restored builder still agrees with the current library API, not just with the 0.9.0 one it was written against.

Collection of the full lane is intact:

```
$ pytest --collect-only -q -m full tests/test_qi_sheet_gate.py
6 tests collected
```

Pre-push gates: `ruff check vmex tests examples benchmarks tools` — all checks passed; `pytest -q tests/test_cited_paths.py tests/test_test_manifest.py` — 33 passed (ESSOS present locally).

The four remaining gate tests are the multi-minute ladder solves the weekly lane exists to run; they are left to CI rather than re-run locally.


rogeriojorge:
Weekly `full-lanes` dispatched on this branch: [run 35442538457](https://github.com/uwplasma/vmex/actions/runs/35442538457) — **Weekly qi-gate: 6 passed in 12:33** (was 1 passed / 5 errors in 59 s on `main`).

```
265.81s setup  tests/test_qi_sheet_gate.py::test_qi_sheet_free_boundary_converges
190.00s call   tests/test_qi_sheet_gate.py::test_qi_sheet_gate_ladder_delt_055_regression
152.11s call   tests/test_qi_sheet_gate.py::test_qi_sheet_gate_ladder_fft_matches_dense
 71.30s call   tests/test_qi_sheet_gate.py::test_qi_sheet_gate_ladder_matches_vmec2000
 44.67s call   tests/test_qi_sheet_gate.py::test_qi_fixed_238_ladder_converges
 27.51s call   tests/test_qi_sheet_gate.py::test_qi_sheet_free_boundary_converges
6 passed in 753.10s (0:12:33)
```

The 265.81 s module-scoped setup is the restored builder constructing the sheet field once, on x86-linux. All five VMEC2000-golden comparisons pass against it — including the DELT = 0.55 regression — so the restored file reproduces the field the goldens were recorded on, not merely an importable module.

The workflow's `job` input is `all` / `campaign` / `full-lanes` with no per-lane choice, so the whole `full-lanes` matrix was dispatched; only `qi-gate` is relevant here.

PR #370, opened 2026-09-19
Stop the 537-mode FFT spy re-declaring _make_body's keywords

rogeriojorge:
## Failing job

[Weekly hmfb](https://github.com/uwplasma/vmex/actions/runs/35425350853/job/105850215439) on `main` (weekly run 35425350853): **1 failed, 3 passed in 1:03:08**.

```
        new_carry = _make_body(
            replace(rt_vac, bsqvac_edge=bsqvac), use_fft=use_fft,
            evaluation_synthesis=(coeffs, geometry, jacobian),
        )(c)
E       TypeError: test_free_boundary_537_modes_fft_auto_smoke.<locals>.recording_body()
        got an unexpected keyword argument 'evaluation_synthesis'

vmex/core/freeboundary.py:1569: TypeError
```

Cost of the stale line: the TypeError was raised 480 s into the solve, from inside the traced steady vacuum lane, in a 63-minute job.

## Cause

Test-side only; no library or physics defect. The `_make_body` spy in `test_free_boundary_537_modes_fft_auto_smoke` re-declared the library's parameter list:

```python
def recording_body(rt, *, evaluation_state=None, use_fft=False):
```

The library signature is now

```
_make_body(rt, *, evaluation_state=None, use_fft=False, evaluation_synthesis=None)
```

`evaluation_synthesis` was added when the steady vacuum lane started computing geometry and its Jacobian once per pass and handing that synthesis to the force evaluation (the `funct3d.f` ordering) instead of synthesising a second time. `freeboundary.py:1569` passes it; the double rejects it.

The sibling test in the same file, `test_use_fft_reaches_every_free_boundary_lane`, already hit this class of failure and carries the comment "`**kwargs`, not the argument list of the day". That lesson was never applied here. Sweeping the whole file: those two are its only doubles of these functions — `_resolve_use_fft` is spied in this test alone, `_make_body` in these two, and the third `monkeypatch.setattr` is a plain value (`_VACUUM_EXECUTABLE_CACHE = {}`).

Nothing caught it at pull-request time because this is a `full`-marked, 8-minute test that only the weekly `hmfb` lane (added in #355) runs.

## Fix

- `recording_body` takes `(rt, *, use_fft=False, **kwargs)` and forwards `**kwargs` — the shape its sibling already uses, so a future keyword cannot break it.
- `recording_resolve` gets the same treatment (`*args, **kwargs`) so a new `_resolve_use_fft` argument cannot repeat this.
- Since the double now sees `evaluation_synthesis`, record the steady-vacuum-lane traces separately and assert the lane was entered.

## Does it still test what its name claims?

The last point matters: with only `assert seen and all(seen)`, the eqsolve lane alone would satisfy the test, and the 537-mode *vacuum* body — the thing that just broke — could go untraced. The new `assert steady` closes that. The name's other two claims are checked directly:

```
mnmax: 537    threshold GPU_MAX_SPECTRAL_MODES: 512
auto _resolve_use_fft(None, ...) -> True
library _make_body signature: (rt, *, evaluation_state=None, use_fft=False,
                               evaluation_synthesis=None) -> Callable[...]
```

537 > 512, so the automatic rule — consulted with `use_fft=None`, which the in-test `assert resolved == [(None, True)]` pins — resolves to FFT. The test also still asserts vacuum activation was reached and the final residual is finite and below 1.

## Local verification

macOS arm64, JAX 0.11.1, absolute `PYTHONPATH` to the worktree (`vmex.__file__` resolves to the worktree, not the installed wheel), heavy-run lock held:

```
$ JAX_ENABLE_X64=1 RUN_FULL=1 pytest -q \
    tests/test_high_mode_free_boundary_parity.py::test_free_boundary_537_modes_fft_auto_smoke
.                                                                        [100%]
============================= slowest 5 durations ==============================
771.49s call  tests/test_high_mode_free_boundary_parity.py::test_free_boundary_537_modes_fft_auto_smoke
1 passed in 771.83s (0:12:51)
```

The full solve was run, not a reduced subset: 12:51 locally against 8:00 on the hosted runner. It passes with the new `assert steady` in place, so the steady vacuum lane really was traced with the FFT body.

Pre-push gates: `ruff check vmex tests examples benchmarks tools` — all checks passed; `pytest -q tests/test_cited_paths.py tests/test_test_manifest.py` — 33 passed (ESSOS present locally).

The other three tests in this lane passed on the failing run and are untouched here.

## Related

The same weekly run's `qi-gate` failure is a separate root cause, fixed in #369.


rogeriojorge:
Weekly `full-lanes` dispatched on this branch: [run 35442546355](https://github.com/uwplasma/vmex/actions/runs/35442546355) — **Weekly hmfb: 4 passed in 1:32:54** (was 1 failed / 3 passed in 1:03:08 on `main`).

```
3704.50s call  tests/..._parity.py::test_free_boundary_238_mode_ladder_survives_activation
 827.15s call  tests/..._parity.py::test_free_boundary_537_modes_fft_auto_smoke
 674.47s call  tests/..._parity.py::test_vacuum_survives_a_radial_transition
 368.24s call  tests/..._parity.py::test_dense_and_fft_free_boundary_trajectories_match
4 passed in 5574.59s (1:32:54)
```

The 537-mode smoke now runs to completion on the hosted runner — 827 s against the 480 s at which it previously died — and passes with `assert steady` in place, so the steady vacuum lane was traced with the FFT body rather than the test passing on the eqsolve lane alone. The other three tests are unchanged and still pass.

The workflow's `job` input is `all` / `campaign` / `full-lanes` with no per-lane choice, so the whole `full-lanes` matrix was dispatched; only `hmfb` is relevant here.

PR #371, opened 2026-09-19
Single stage: three self-contained scripts, adding a least-squares form

rogeriojorge:
Follows simsopt's `examples/2_Intermediate/stage_two_optimization.py` convention. Each script stands alone — a reader never has to open a second file to follow one workflow, and can copy one and edit it.

| script | constraints | optimizer |
| --- | --- | --- |
| `single_stage_optimization.py` | quadratic penalties | L-BFGS-B |
| `single_stage_optimization_auglag.py` | Powell-Hestenes-Rockafellar multipliers | L-BFGS-B over stages |
| `single_stage_optimization_least_squares.py` *(new)* | one-sided residuals | TRF Gauss-Newton |

## Structure

Each file runs: docstring stating the objective → flat imports from VMEX and ESSOS → a labelled block of input parameters → an explicit `End of input parameters` banner → then sections for the equilibrium, the coils, the objective, the optimizer, the run, and printing/plotting/saving.

An earlier pass factored the shared setup and the verify/save/plot tail into a `_single_stage.py` module, which cut each script from ~490 to ~190 lines. **That is reverted.** Line count was the wrong target for a teaching example: the tail it hid — the finer verification solve, the explicit target checks, the saving and plotting — is part of what the script is there to teach, and splitting it forces a reader between two files. The scripts are 447, 509 and 462 lines and read top to bottom.

## The least-squares form

It keeps the terms as a residual vector and hands SciPy a Jacobian, so a Gauss-Newton step uses the curvature that `(1/2) rᵀr` implies. Two things the file documents because they are easy to trip on:

- **The joint Jacobian is cheap.** Plasma residuals depend only on the boundary coefficients and the coil residuals are pure JAX, so the block coupling plasma rows to coil dofs is *identically zero* and only the boundary columns need an implicit equilibrium solve.
- **A guarded `sqrt` is required.** The two clearance terms are aggregate hinge losses, so a residual row is their square root — and that row is exactly zero once the clearance is met, which is where `sqrt` has an infinite derivative. Without the doubled-`where` guard the very first Jacobian comes back NaN.

Constraint handling is unchanged from the penalized form, and for a structural reason: least squares minimizes a sum of squares, and a one-sided residual squared *is* a quadratic penalty. What differs is the step, not the formulation.

## Verified

All three run end to end in smoke mode, exit 0 with the objective falling. The L-BFGS-B script reproduces the merged penalty script's objective trace exactly (2.276182e+04 → 7.371648e+03), so the rewrite is behaviourally identical to what shipped in #368. `tools/preflight.py --static` passes, including the example-coverage and manifest guards; tests added for the two new scripts.

## Not verified, and deliberately left

- **No full-budget runs since the rewrite.** Before it, at coil length 5.3 and coil–coil 0.17, L-BFGS-B took 697 s and the augmented Lagrangian 755 s, both meeting every target. The least-squares form has never been run to full budget, so whether it converges in fewer trials — the reason to prefer it — is untested.
- The finite-beta and the two free-boundary single-stage examples (340, 245 and 198 lines) still need the same treatment.


rogeriojorge:
Superseded by #411, which contains this split with the spelled-out augmented-Lagrangian name and the matched fixed/free-boundary examples.

PR #411, opened 2026-09-21
Single stage: simple name for the simple script, spelled-out augmented Lagrangian, least-squares form

rogeriojorge:
**Status: handoff, not ready to merge.** The renames and reference updates are done and the scripts pass their smoke runs. What's missing is the full-budget measurement, so the docstrings' comparison numbers are not yet re-verified. This supersedes #371 and carries its content over. Close #371 when this merges.

## Rename table

| before (main) | after | form |
| --- | --- | --- |
| `single_stage_optimization_penalty.py` | `single_stage_optimization.py` | quadratic penalties, one bounded L-BFGS-B solve. Read this one first. |
| `single_stage_optimization.py` | `single_stage_optimization_augmented_lagrangian.py` | Powell-Hestenes-Rockafellar multipliers, L-BFGS-B per stage |
| (from #371) | `single_stage_optimization_least_squares.py` | one-sided residuals, bounded TRF Gauss-Newton |

The first commit contains only the two `git mv`s, so the swap is visible as renames. Because two files trade places, git's similarity detection pairs them however it likes; read commit 1 for the actual moves. Output files follow the script names:
- penalized: `single_stage_optimization.png`, `single_stage_optimization_summary.json`, `wout_single_stage_optimized.nc`, and so on
- augmented Lagrangian: `single_stage_augmented_lagrangian_*`
- least squares: `single_stage_least_squares_*`

## Done

- #371's rewrite is brought in. It uses the same layout in all three files, has no shared module, and keeps every example as one flat script with no argparse and no `__main__`. Input parameters are identical to main's, checked line by line.
- Defects in #371, fixed here:
  - The augmented-Lagrangian docstring compared the file with itself, and the least-squares docstring described itself as the scalar form. Both now name `single_stage_optimization.py`.
  - All three scripts wrote `single_stage_optimization.png`/`.gif`, so each run overwrote the others' figures.
  - The least-squares monitor recorded only the starting point (no callback was passed to `least_squares`), so its CSV, plot and movie were empty. It now passes `callback=monitor`, which SciPy has had since 1.16 (the repo's floor), with per-block terms and `J.T r` as the gradient.
  - `MOVIE_SURFACE_COLOR` is restored, and the multiplier helper is renamed `augmented_lagrangian`. Both are asserted by existing static tests that #371 left failing.
- References updated:
  - `tests/test_examples.py`: coil-example tuple, declared-example set, the two nightly run tests plus a new least-squares one, and the static contract test, which now checks all three files.
  - `examples/README.md`
  - `benchmarks/INDEX.md`: the 2026-09-13 profile was of the augmented-Lagrangian script, so its row now names that file.
  - `git grep` finds no hits for `_penalty.py`, `_auglag`, `single_stage_penalty_` or `single_stage_auglag_`.
  - The root `README.md` is not touched (another agent owns it). Its references to `single_stage_optimization.py` still resolve, and now point to the penalized form.
- Removed by the #371 rewrite relative to main: the `VARY_MAJOR_RADIUS` knob, the `dof_names` bookkeeping, the seed-history comment, and the per-term helper functions, which were folded into the two objective functions.

## Measured

- Smoke mode (`VMEX_EXAMPLES_CI=1`), all three:
  - Penalized: objective 2.276182e+04 -> 7.371648e+03, identical to main's penalty script.
  - Augmented Lagrangian: 3.382159e+02 -> 2.283141e+02 over 1 stage.
  - Least squares: 2.275967e+04 -> 8.995810e+03 in 4 trials.
- `pytest tests/test_examples.py -m "not full"`: 43 passed.
- **Blocker on main, fixed separately in #409:** since #403, `surface_field_data_from_state` raises `TypeError: ... unexpected keyword argument 'lmns'`. All three scripts therefore crash at the movie step, after the summary JSON has been written. Merge #409 first. With #409 applied, the least-squares smoke run exits 0.

## Not measured, still to do

1. **Full-budget runs of all three.** The docstring numbers are from before the rewrite: at coil length 5.3 and coil-coil 0.17, L-BFGS-B took 697 s and 149 iterations and the augmented Lagrangian 755 s and 127 iterations (6 stages), both at 301 trials and both meeting every target. The augmented-Lagrangian docstring's "1.4x lower objective" (4.49e-02 against 6.48e-02) comes from those runs. Re-run each script without `VMEX_EXAMPLES_CI`, from a pinned clean checkout that includes #409, one at a time and alternating order, and record `*_summary.json` (`met`, `trials`, `equilibrium_solves`, `optimization_seconds`). Then update or remove the numbers in the docstrings.
2. **Least-squares trial count.** This has never been run to full budget. Each trial costs one implicit solve per boundary column instead of one adjoint. Its docstring claims only that "a step should go further". Either measure that, or cut the claim.
3. **Free-boundary augmented Lagrangian: not decided and nothing added.** One finding so far: the "status 2 at excursion 1e-1" in `single_stage_free_boundary_optimization.py` comes from *seeding* experiments (random excursions from circular coils), not from optimizer iterates. The optimizer starts from LandremanPaulQA coils with min |iota| 0.4137, and its only constraint is the iota floor. An augmented Lagrangian changes how constraints pull on iterates, not the seed, so that failure mode is not evidence for adding one. The next step is to run the existing script at full budget and count rejected trials (`status != 0`) and free-boundary solves. Add a scratch augmented-Lagrangian variant only if that run shows rejected trials or a missed iota floor. Otherwise record the numbers and leave the free-boundary examples alone.
4. `benchmarks/single_stage_profile.py` runs the example by basename (`"fixed": "single_stage_optimization"`), so re-running it now profiles the penalized form rather than the augmented Lagrangian its committed JSON came from. Point it at `single_stage_optimization_augmented_lagrangian` if the committed profile is ever regenerated.
5. `plan.md` is not updated. Its single-stage entries describe the old names historically.
6. When everything is green, merge with `gh pr merge <n> --squash --admin --delete-branch`, then close #371 with a pointer here.


rogeriojorge:
Measured runs at `e0fa21745` (clearance limit 0.15 m, coil order 3, MAXITER 50 / MAX_TRIALS 100). All four ran concurrently on one laptop, so wall times are pessimistic.

| example | wall | peak RSS | iterations / trials | stop reason | targets |
| fixed, zero beta | 477 s | 3.2 GB | 50 / 57 | iteration cap | misses aspect only: 4.0206 vs <= 4.0 (min iota 0.4252, B.n RMS 0.803%, clearance 0.221 m) |
| fixed, 0.5% beta | 844 s | 11.9 GB | 50 / 63 | iteration cap | **all met** (beta 0.4999%, min iota 0.4283, aspect 3.9939, B.n RMS 0.802%) |
| free, zero beta | 1839 s | 6.5 GB | 13 / 42 (5 rejected) | factr (objective 2.75 -> 0.0717) | misses min iota 0.4099 vs 0.42, B.n RMS 3.57% vs 1%, clearance 0.1325 vs 0.15; verification free-boundary solve did not converge |
| free, 0.5% beta | 1981 s | 10.4 GB | 35 / 67 (4 rejected) | factr (objective 1.97 -> 0.0227) | misses min iota 0.3426 vs 0.42, beta 0.435% vs 0.45%; verification free-boundary solve did not converge |

Findings for whoever picks this up:
1. Free boundary: the optimizer reports convergence with a small objective while the independent final check misses several targets and its verification solve fails. The objective's terms and the final check disagree (this is also the earlier 0.21 m vs 0.153 m clearance discrepancy). Suspect the per-trial boundary helper / the surface the coil terms are evaluated on. Fix that before tuning weights or budgets.
2. Free-boundary cost is per-trial (~27-44 s under load), not iteration count: both runs stopped well before 50 iterations and still took ~30 min.
3. QA total rises in every run (free zero beta 0.098 -> 0.289, free 0.5% 0.114 -> 0.535, fixed 0.101 -> 0.118): coil terms dominate the weighted objective.
4. Fixed zero beta misses aspect by 0.5% at the iteration cap; fixed 0.5% beta meets everything in 50 iterations.

rogeriojorge:
Runs at `952c31601` (aspect limit 6.0 with an aspect-5.97 seed, coil order 4, coils on a 0.5 m circle, one-sided coil length <= 5.5 m at weight 0.5, clearance 0.15 m, 50 iterations / 100 trials). Four runs concurrent on one laptop, so wall times are pessimistic.

| example | targets | wall | peak RSS | iterations / trials |
| fixed, zero beta | **all met** (min iota 0.4289, aspect 5.969, B.n RMS 0.800%, clearance 0.261 m) | 335 s | 3.8 GB | 50 / 86 |
| fixed, 0.5% beta | **all met** (beta 0.5011%, min iota 0.4295, aspect 5.971, B.n RMS 0.800%) | 566 s | 9.1 GB | 50 / 65 |
| free, 0.5% beta | **all met** (beta 0.4833%, min iota 0.4309, aspect 5.930, clearance 0.246 m) | 2399 s | 14.3 GB | 27 / 109 (trial cap) |
| free, zero beta | misses B.n RMS 1.48% vs 1%; verification free-boundary solve did not converge | 1536 s | 7.7 GB | 27 / 95 (9 rejected) |

Findings:
1. Coil terms are no longer active: length, curvature, coil-coil and coil-surface terms are ~0 at the end of every run (coils 4.5-4.8 m).
2. Quasisymmetry is traded against the iota floor, not coils: the seed starts at min|iota| 0.404 < IOTA_CONSTRAINT 0.43 and the iota/aspect penalty is 1.87 of the 1.90 starting objective (fixed, zero beta); QS goes 0.0265 -> 0.0281. For QS to improve, start the seed above the floor or lower the floor to ~0.40.
3. **Bug, free boundary zero beta:** L-BFGS-B reports status 0 while the objective goes 1.79 -> 4155 and every logged term is 0 at the end, i.e. the returned/reported point is a rejected trial carrying the failure penalty rather than the best accepted iterate. Report and verify the best accepted iterate, and investigate why the verification free-boundary solve fails there.
4. Free-boundary wall time (26-40 min) is still too long for a user-facing example; per-trial cost ~16-22 s vs ~4-9 s fixed boundary.
5. Docstring numbers in the four scripts predate these runs and must be refreshed from this table before merging.

rogeriojorge:
Free boundary, zero beta, at `8bab51c14`: the best-accepted-trial fix changes only the reported objective (now 1.79 -> 2.275e-2 instead of the 4155 wall value). SciPy's `result.x` was already the best accepted iterate, so the verification ran at the right point and still did not converge.

Replay of only the verification solve at the optimized coils (cold start, `solve_free_boundary_multigrid`):

| ns | ftol | niter | final fsq (r, z, l) | converged | B.n/B RMS | time |
| --- | --- | --- | --- | --- | --- | --- |
| 16 -> 51 (shipped) | 1e-12 | 8000 | 1.7e-5 at the last step, oscillating from ~1e-8 | no | 1.48% | - |
| 31 | 1e-10 | 8000 | 6.8e-8, 1.4e-8, 3.7e-9 | no | 1.06% | 33 s |
| 31 | 1e-12 | 20000 | 4.1e-9, 1.9e-9, 2.7e-10 | no | 0.85% | 62 s |
| 31 | 1e-10 | 40000 | 1.3e-8, 4.9e-9, 8.7e-10 | no | 0.94% | 117 s |
| 31 | 1e-11 | 60000 | 4.1e-9, 2.0e-9, 8.6e-10 | no | 0.80% | 169 s |
| 16 -> 31 | 1e-11 | 10000 | 7.7e-7, 4.4e-7, 4.5e-8 | no | 0.85% | 65 s |
| 31 -> 51 | 1e-11 | 10000 | 4.5e-7, 5.3e-8, 5.6e-9 | no | 1.19% | 76 s |

Conclusions:
- Not an iteration budget: at ns = 31 the residual floors at ~4e-9 to 1.3e-8 and oscillates (20k, 40k and 60k iterations end at 4.1e-9, 1.3e-8, 4.1e-9, depending on the phase of the cycle). ns = 16 and ns = 51 are worse. More ns or more niter during the optimization would not change this.
- The finite-beta free-boundary verification converges cleanly (1e-12 at ns = 16 and 51 in under 2000 iterations each, DELT steady), so this is specific to the zero-beta (vacuum) free boundary.
- The physics target is met once the boundary is resolved to ~1e-9: B.n/B RMS 0.80-0.85% (target 1%). The 1.48% came from the unconverged 16 -> 51 ladder.
- Open solver question: during optimization this point was accepted as status 0 at ns = 31, ftol 1e-10 (warm-started from the previous trial), while a cold start never gets below ~4e-9 in 60000 iterations. Either the warm start genuinely reaches the root, or the implicit path's acceptance is looser than ftol. Check the fsq recorded for the accepted trial before deciding.
- Options for the example: (a) verify at ns = 31, 20000 iterations, and state the plateau (~1e-8) as the achievable tolerance for this vacuum case; (b) warm-start the verification from the optimizer's accepted state; (c) fix the vacuum limit cycle in the solver. (a) weakens a certificate and should be an explicit maintainer decision.

rogeriojorge:
Warm-started verification at `08d97509f` (free boundary, zero beta): **does not fix it.**

- Re-solving the optimizer's accepted point with `solve_free_boundary_implicit_status` (ns = 31, ftol 1e-10) now returns **status 2, fsq = 1.374e-8** (ratio ~137). During the optimization the same coil vector was accepted as status 0. The trial status at a fixed point therefore depends on the warm start the solve inherits, i.e. on optimizer history. This is a correctness defect in the same class as #383/#416: gradients were taken at a point that is not reproducibly converged.
- Continuing from that state to ftol 1e-12 at ns = 31 ends at fsq ~2e-7 after 8000 iterations, oscillating with DELT shrinking to 0.26, the same limit cycle as the cold starts. B.n/B RMS 1.18% (target 1%). All other targets pass (min iota 0.444, aspect 5.83, clearance 0.231 m).
- Likely physics: at zero beta and zero current every nested vacuum flux surface is an equilibrium, so the free boundary is fixed only through PHIEDGE and is weakly determined; the NESTOR iteration drifts along that near-null direction. The 0.5% beta example, with the same coils and settings, converges cold to 1e-12 at ns = 16 and 51 in under 2000 iterations.

Next steps for whoever picks this up:
1. Solver: reproduce status 0 vs 2 at the same coil vector from two different warm starts, and decide whether the implicit status must be computed from a start-independent state (as #383 did for the forward solve).
2. Example: either drop the zero-beta free-boundary example in favour of the 0.5% beta one, or make its FTOL the reproducible level (~1e-8) and say so. The warm-start commit can be reverted (it removes the independence of the check without making it converge); keep the `[accepted]` diagnostic line.

rogeriojorge:
## Deferred correctness work: the VMEX free-boundary solve in vacuum (zero beta, zero current)

Owner's decision: fix this in the full free-boundary solve itself, as a real correctness fix for vacuum free boundaries, not by working around it in the example. It is to be researched and fixed later; this comment records everything found so far so the work can start cold.

### What is wrong

1. **Trial status is history-dependent at zero beta.** The zero-beta single-stage optimization accepted the final coil vector as status 0 (converged, derivative-certified) at ns = 31, ftol 1e-10. Re-solving the *same* coil vector with `solve_free_boundary_implicit_status` returns **status 2, fsq = 1.374e-8** (ratio ~137 against ftol). Whether a trial counts as converged depends on the warm start it inherits from the optimizer's history, so gradients were taken at a point that is not reproducibly an equilibrium. Same defect class as #383 (history-dependent forward roots) and #416 (history-dependent cold recovery).
2. **The vacuum free boundary does not converge; it limit-cycles.** At the optimized coils, the force residual bottoms out near 1e-9 to 1e-8 and then oscillates, bursting back to 1e-5 to 1e-4 every 200-400 iterations while DELT shrinks (0.50 -> 0.26). No iteration budget or radial resolution fixes it (cold start, `solve_free_boundary_multigrid`):

| ns | ftol | niter | final fsq (r, z, l) | converged | B.n/B RMS |
| 16 -> 51 | 1e-12 | 8000 each | ~1e-8, last 1.7e-5 | no | 1.48% |
| 31 | 1e-10 | 8000 | 6.8e-8, 1.4e-8, 3.7e-9 | no | 1.06% |
| 31 | 1e-12 | 20000 | 4.1e-9, 1.9e-9, 2.7e-10 | no | 0.85% |
| 31 | 1e-10 | 40000 | 1.3e-8, 4.9e-9, 8.7e-10 | no | 0.94% |
| 31 | 1e-11 | 60000 | 4.1e-9, 2.0e-9, 8.6e-10 | no | 0.80% |
| 16 -> 31 | 1e-11 | 10000 | 7.7e-7, 4.4e-7, 4.5e-8 | no | 0.85% |
| 31 -> 51 | 1e-11 | 10000 | 4.5e-7, 5.3e-8, 5.6e-9 | no | 1.19% |
| 31, warm-started from the accepted state | 1e-12 | 8000 | ~2e-7 at the end, oscillating | no | 1.18% |

   The final value at ns = 31 depends on the phase of the cycle (20k, 40k, 60k iterations end at 4.1e-9, 1.3e-8, 4.1e-9), not on the budget.
3. **Finite beta is fine.** The 0.5% beta free-boundary example with the same seed, coils and settings converges from a cold start to 1e-12 at ns = 16 and ns = 51 in under 2000 iterations each, DELT steady at 0.45-0.50, and meets every target.

### Likely cause (hypothesis, to be verified)

With zero pressure and zero net current, every nested flux surface of the vacuum field is itself an equilibrium. The free boundary is then selected only by the enclosed toroidal flux (PHIEDGE), so the coupled plasma/NESTOR problem has a near-null direction: moving the boundary to a neighbouring vacuum flux surface changes the force residual only weakly. The iteration drifts along it and the vacuum-pressure coupling produces the limit cycle. Pressure (finite beta) removes the degeneracy, which matches the clean finite-beta convergence.

### Directions to research

- **Small-beta limit.** Solve the vacuum case as the limit of a very small beta (e.g. 1e-6 to 1e-4 with the same simple pressure profile), and establish whether (a) it converges cold and reproducibly, (b) the result converges to the vacuum answer as beta -> 0 (B.n/B, iota, boundary shape), and (c) what the smallest beta is that gives a well-conditioned solve. If this works, VMEX could use it internally for vacuum free boundaries (a regularization with a stated error), rather than asking users to do it.
- **Fix the vacuum iteration directly.** Identify the near-null mode (e.g. from the linearized free-boundary operator at the optimized coils) and constrain or damp it, or change the vacuum/plasma coupling schedule (NESTOR update frequency, DELT control) so it converges instead of cycling. Compare with VMEC2000 on the same coils and PHIEDGE to see whether the Fortran code converges this case.
- **Make trial status start-independent.** Whatever the forward fix, `solve_free_boundary_implicit_status` must return the same status and state for the same coil vector regardless of the optimizer's history (as #383 did for the forward solve). Add a regression that evaluates one zero-beta point from two different warm starts and requires identical status.
- **Regression case.** The optimized coils and deck from this run reproduce the cycle in ~30-170 s per replay: `examples/optimization/single_stage_free_boundary_optimization.py` at this branch's head, then replay the verification solve from the written `input.single_stage_free_boundary_optimized` and `coils_single_stage_free_boundary_optimized.json`.

### For now (this PR)

- The free-boundary scripts report and verify the best accepted trial (commit `8bab51c14`); that only corrected the reported objective (4155 -> 2.275e-2), since SciPy's returned point was already the best accepted one.
- The zero-beta verification is warm-started from the accepted state and prints the accepted trial's recorded fsq (commit `08d97509f`). The warm start does not make it converge and removes the check's independence; revert it when the solver is fixed, keeping the `[accepted]` diagnostic line.
- Current results at this branch: fixed zero beta, fixed 0.5% beta and free 0.5% beta meet every target; free zero beta meets every target except B.n/B RMS (1.18% vs 1%) and its verification solve, both consequences of the vacuum limit cycle above.

rogeriojorge:
## Diagnosis: the zero-beta free boundary does not converge because the target flux sits on a 4/9 island chain of the coil field

Follow-up to "Deferred correctness work: the VMEX free-boundary solve in vacuum". Short version: **this is not a VMEX solver defect, and not a zero-beta degeneracy.** The optimized coils make a vacuum field whose 4/9 island chain, with a stochastic layer outside it, sits exactly at the requested enclosed flux. VMEC's nested-surface model has no solution there. VMEC2000 and VMEC++ fail on the same case in the same way. Adding an upper bound on iota to the example keeps the resonance out of the plasma, and then the cold 16 -> 51 verification converges to 1e-12 and every target is met.

All timings below were measured on a laptop running other heavy jobs. Unless stated otherwise, runs are cold starts from the deck's rotating-ellipse seed, NFP 2, MPOL = NTOR = 5, NTHETA 16, NZETA 14, with the direct ESSOS field.

### 1. Field-line tracing of the optimized coils (H1: confirmed)

I traced 64 field lines from the magnetic axis outwards along the outboard midplane (phi = 0), for 1000 toroidal transits each (RK4, 160 steps per field period). The toroidal flux inside each traced surface was integrated on the phi = 0 plane.

| start (fraction of axis-to-boundary distance) | field-line iota | enclosed flux / PHIEDGE |
| --- | --- | --- |
| 0.06 - 0.79 | 0.4594 -> 0.4493 (nested surfaces, low shear) | up to **0.79** (last intact surface) |
| 0.81 - 0.83 | separatrix layer | ~1.0 |
| 0.85 - 1.03 | **0.44433 - 0.44446, locked at 4/9 = 0.44444**: a 9-island chain | the islands |
| 1.05 - 1.40 | stochastic layer; most lines started beyond 1.09 are lost | |

The Poincare sections at phi = 0, pi/4 and pi/2 show a 9-island chain. The PHIEDGE boundary from VMEX (full flux) cuts through the islands and, at phi = pi/4, reaches into the stochastic layer. **The last closed flux surface of this coil field encloses about 0.79 x PHIEDGE.** No nested flux surface encloses the requested flux, so a zero-beta, zero-current equilibrium with nested surfaces does not exist for this deck.

Two consequences follow. The residual cannot go to zero: the boundary is not a flux surface of the coil field, and B.n/B of the coil field stays at 0.8 to 1.5% while DEL-BSQ stays at about 7e-3. And the reported state depends on where the iteration happens to be in its cycle, which is why the trial status was history-dependent.

A secondary effect: while VMEX is not converged, its iota (edge 0.426 to 0.44) reads lower than the true field-line iota (0.444 to 0.459). The `min |iota| >= 0.43` penalty therefore pushed the real profile, which falls slightly outwards, up through 4/9. The seed profile, 0.405 to 0.41, was just above 2/5.

### 2. The same deck in VMEC2000 and VMEC++ (the fault is not in VMEX)

I tabulated the same coils on an mgrid (181 x 181 x 14, R in [0.55, 1.45], Z in [-0.45, 0.45]) and used PHIEDGE of the same physical flux; the sign convention differs.

| code | ns | result |
| --- | --- | --- |
| VMEC2000 | 31 | no convergence. fsq floors at 3.6e-8 and bursts to 1.7e-4 and 3.8e-4. The run keeps restarting to the same stored state while DELT collapses from 0.50 to 7.4e-3 (stopped at about 9600 iterations). |
| VMEC++ | 31 | no convergence. Bursts to 4e-5 while DELT falls from 0.50 to 0.04, then fatal `JACOBIAN_75_TIMES_BAD` at 9022 iterations. |
| VMEX | 31 | no convergence. Floor about 4e-9, bursts to 1e-5 to 1e-4, DELT 0.50 -> 0.30 (as reported before). |

### 3. Control: vacuum free boundary works when nested surfaces exist

These use the Landreman-Paul QA ESSOS coils from `examples/data`, zero beta, zero current, PHIEDGE -0.025, same MPOL/NTOR, direct field.

- Cold ns = 31: converged to 1e-11 in 1832 iterations.
- Cold 16 -> 51: converged in 627 iterations at ns = 51.

The earlier hypothesis was that at zero beta every vacuum surface is an equilibrium and this leaves a near-null direction. That is ruled out: the same solver, at zero beta, converges promptly on a field that has good surfaces at the requested flux.

### 4. Other hypotheses, tested and ruled out

- **H3, small-beta regularization** (p = p0 (1 - s), zero current, cold, ns = 31, 20000 iterations): none converged.

  | <beta> | 1.0e-6 | 1.0e-5 | 9.9e-5 | 1.0e-3 | 5.4e-3 |
  | final fsqr | 2.1e-7 | 7.3e-8 | 1.8e-5 | 4.0e-7 | 2.5e-6 |

  Pressure does not create the missing flux surface. The 0.5% beta example converges because its own optimized coils are different, not because beta regularizes the problem. **Tiny beta is not a valid internal regularization for this case**, and I do not recommend building it into VMEX.
- **H2, NESTOR cadence and step size**: NVACSKIP 6 gives fsqr 1.7e-6, NVACSKIP 15 gives 2.8e-7 and DELT 0.3 gives 3.2e-6, all at 20000 iterations and none converged. Raising the resolution does not help either (MPOL = NTOR = 7 at PHIEDGE x 0.7: 2.3e-7).
- **H4, a code path special to zero pressure**: none found. The edge pressure term `presf_ns_scale` already returns 0 when p(1) = 0, and it is 0 for the finite-beta p = p0 (1 - s) as well. The other free-boundary paths (activation, `ivac`, the `nvacskip` cadence, the `medge` rule) do not depend on pressure and follow VMEC2000.

### 5. Inside the good surfaces a vacuum solution exists, but the ellipse seed is a poor starting point

| PHIEDGE scale | seed | result |
| --- | --- | --- |
| 0.85 | ellipse | not converged at ns 31 or 16 -> 51 (the flux is in the separatrix region) |
| 0.70 | ellipse | 16 -> 51 converged in 1075 iterations; ns 31 and ns 51 alone did not |
| 0.40, 0.55, 0.60 | ellipse (also scaled to the smaller flux) | not converged |
| 0.40 | interior flux surface of the converged 0.70 equilibrium | **converged, ns = 31, 814 iterations** |
| 0.55 | same | fsq 1.1e-11 against ftol 1e-11 (stalled at tolerance) |
| 0.70 | converged 0.70 boundary | converged, ns = 31, 975 iterations |

In the converged states, iota matches the field-line tracing to within about 0.005 (0.4615 on axis against 0.459 traced). From the rotating-ellipse seed, VMEC2000 and VMEC++ also fail at PHIEDGE x 0.70. VMEC++ does not converge at ns = 16 in 10000 iterations, and VMEC2000 was still at 3.7e-4 when I stopped it after 1800 iterations. VMEX converged that case only with the 16 -> 51 ladder. This matches Hirshman, van Rij & Merkel (1986), whose vacuum test reproduced "a magnetic surface of the analytic vacuum field" only for a field with "a large domain of 'good' magnetic surfaces".

### 6. What the example should do: an iota ceiling below 4/9 (measured)

I added a one-sided upper bound on iota next to the existing floor: target `max |iota| <= 0.44`, penalty at 0.437, built as `jnp.max(jnp.abs(im.iota_profile(state, ctx)[1:]))` and fed into the same hinge. I also reverted the warm-started verification (commit `08d97509f`) to a cold 16 -> 51 ladder at ftol 1e-12. I then ran the full example with nothing else changed. Its wall time was 1898 s under load.

- Optimization: 30 L-BFGS-B iterations, 101 trials, **2 rejected** (the previous run had 9). Objective 1.79 -> 2.72e-2. QA total 0.0551, against 0.0545 before.
- **Cold verification converged**: ns = 16 reached 1e-12 in 1738 iterations and ns = 51 in 2454, with DELT steady at 0.45 to 0.50 and no bursts.
- Every target is met: min |iota| 0.431, max |iota| 0.436, aspect 5.93, **B.n/B RMS 0.617%** (target 1%, was 1.48%), coil-surface distance 0.244 m, coil-coil distance 0.191 m, maximum curvature 6.90.
- Reproducibility on the new coils: independent cold solves at ns = 16, 31 and 51 all converge (1413, 1962 and 2545 iterations), and each gives B.n/B of 0.617 to 0.618%.
- Field-line tracing of the new coils: iota runs from 0.434 to 0.438 across the plasma with no locked chain, and nested surfaces extend to about 1.11 times the boundary's distance from the axis on the outboard midplane. The margin is thin at phi = pi/4.

Recommended change to `single_stage_free_boundary_optimization.py`:

1. Add the iota ceiling. Keeping the whole profile in [0.42, 0.44] leaves out 2/5 and 4/9. The rationals left in that window with n a multiple of NFP are high order: 3/7 (m = 14) and 5/12 (m = 24).
2. Restore the cold 16 -> 51 verification. Keep the `[accepted]` diagnostic line, but update the comment block above it, which still describes the warm start.
3. Add `max |iota|` to the printed report and to the summary targets.

### Solver follow-ups (optional, not fixes)

At zero beta and zero current, a converged free boundary must be a flux surface of the coil field. VMEX could report the coil-field B.n/B RMS on the boundary, or DEL-BSQ, for vacuum free-boundary solves and warn when it does not fall with resolution. That would catch this failure directly, where it now shows up as a limit cycle. The trial status that depends on history is a symptom of the missing equilibrium: at the old coils there is no fixed point for the solve to reach reproducibly.

### References consulted

- S. P. Hirshman, W. I. van Rij, P. Merkel, Comput. Phys. Commun. 43, 143 (1986), NESTOR. Section 4 gives the pressure-balance and B.n = 0 conditions. The vacuum test converges to a magnetic surface in a field with a large domain of good surfaces, and the selected surface depends only on the external field and the toroidal flux.
- DESC free-boundary tutorial (virtual casing, B.n = 0 plus pressure balance, with enclosed flux choosing the surface); R. Conlin et al., "High order free boundary MHD equilibria in DESC" (arXiv:2412.05680). At zero pressure, when a solution matches the coils, the jump in B across the boundary is zero and so is the sheet current.
- J. Loizu, S. R. Hudson et al., Phys. Plasmas 22, 090704 (2015) and Phys. Plasmas 22, 022501 (2015), on ideal equilibria at rational surfaces. Nested-surface equilibria cannot represent island topology. When islands matter, stepped-pressure codes (SPEC) or field-line tracing are the right tools.

rogeriojorge:
## Handoff: example wall-time review (paused on the owner's budget)

Goal: every optimization example should finish within 5 minutes, end to end, on an unloaded laptop. The work stopped partway; this comment records what was measured and what is left.

**Measurement setup.**
- Pinned, clean worktrees: main `07a47279d`, #426 `02fee4012`, #411 `35024f378` and its successor `64d18ed3e`.
- Fresh environment: jax/jaxlib 0.11.2, scipy 1.18, essos 0.17, virtual-casing-jax 0.0.7, booz_xform_jax 0.4.0. Cold JAX compilation cache on every run.
- A wrapper times the public entry points, the implicit-solve counters and JAX compile events. It also saves the accepted iterates, and a second script runs each example's own end-of-run check at chosen iterates.
- **Both machines were saturated by other jobs the whole time.** Laptop (Apple M3 Max, 14 cores): load average 45-107. Office workstation (36-thread Xeon): load 45-107, and it runs about 2x slower than the laptop.
- Wall times below are therefore inflated, and none is an unloaded number. CPU time and the laptop's retired-instruction and cycle counters are more stable. The same fixed-boundary script and budget measured 7.5e12 instructions yesterday and 8.3e12 today, while wall time went from 335 s to 1220 s.
- Unloaded estimates use cycles / 6.7e9 per second. That rate comes from yesterday's 335 s run, which was itself loaded, so the estimates err high.

### Single stage (#411, owned here)

| example | before (35024f378) | breakdown (wall at load) | after (64d18ed3e) | targets |
| fixed, zero beta | 1220 s at load 45-86; 2.92e12 cycles, about 435 s unloaded (est.) | first value+grad compile 229 s (one 108 s XLA compile of the jitted plasma objective); 50 iterations / 90 trials 518 s; **absB movie 233 s**; plot_wout 50 s; ns=101 check 42 s + wout 12 s; VTK 16 s | 12 iterations / 21 trials, movie off: 1252 s at load 45-96, but **1.34e12 cycles, about 200 s unloaded (est.)**, CPU 411 s | all met (min iota 0.4294, aspect 5.970, B.n RMS 0.80%, clearance 0.264 m) |
| fixed, 0.5% beta | 1781 s at load 45-107; 5.34e12 cycles, about 800 s unloaded (est.) | setup 200 s (the seed solve 24 s, then problem build and the seed VC precision plan about 107 s); first value+grad compile 197 s; 50 iterations / 66 trials 849 s; **movie 292 s**; plot_wout 65 s; final VC check on 61x64 about 130 s | 14 iterations, movie off: end-to-end run **not done** (stopped by the pause). Iterate checks: met at 12, 14, 16 and 20; missed at 10 (aspect 6.002) | met at the shipped iterate (14) by the script's own check |
| free, zero beta (iota ceiling) | stopped at 33 min, 16 iterations, about 30 trials, load 60-95 | value+grad ready after 208 s; trials 16-50 s | not changed | not reached |
| free, 0.5% beta | 2399 s yesterday (27 iterations / 109 trials, all met) | compile + first trial 134 s; trials 16 s near the seed, 50 s at 0.1-0.15 away (restart waste) | not changed | - |

Wall to targets, read from saved iterates: fixed zero beta passes every check from iteration 6 on (6, 8, 10, 12, 15, 20 checked). Past about iteration 10 the objective only trades quasisymmetry (0.0295 -> 0.0280 over the next 40 iterations), so the shipped 12 iterations lose nothing visible.

**Pushed (`64d18ed3e`):**
- `MAXITER` 50 -> 12 (fixed, zero beta) and 50 -> 14 (fixed, 0.5% beta), with the evidence in the budget comments.
- `MAKE_MOVIE` now defaults to False in both. With absB colour every frame re-solves its equilibrium, and the movie took longer than the whole shortened optimization.
- The static single-stage tests pass, and ruff is clean.

**Not done:**
- The augmented-Lagrangian and least-squares companions still ship `MAX_TRIALS = 300` and the movie on, and they were not measured.
- Docstring wall times have not been updated, because no unloaded measurement exists yet.

### Free boundary: why it does not fit yet
- A trial is one NESTOR solve plus the boundary-Schur adjoint. On base, every trial restarts from the seed reference and runs to the 4000-iteration cap once the coils have moved away from it, then solves cold. That waste appears at small steps too (50 s vs 16 s here).
- #432 bounds the restart and adds the Newton anchor. Its author measures about 11 s per trial near the seed and 27 s near the optimum on an unloaded-laptop scale. After compile (about 1-2 min) and the [16, 51] verification, that leaves room for about 15 trials.
- Yesterday's finite-beta run cleared its iota/aspect penalty by iteration 5. The cheapest honest route is therefore a small iteration budget, checked with the iterate checker, plus ns 16 for the trial solves, keeping the [16, 51] verification. Neither was measured before the pause. #432 warns that ftol cannot be loosened below 1e-10 with the anchor.

### Other optimization examples (measure-and-recommend; the #426 agent was sent this)
The office queue was stopped after the first QA runs (load 45-107, about 2x laptop):

| example | main | #426 | where the time goes |
| --- | --- | --- | --- |
| QA_optimization.py | 917 s | 2623 s | #426 stage max_mode 5: 1295 s alone. Jacobian 1195 s / 4776 columns, refinement 925 s, compile 551 s, NS=71 certificate 82 s, plot_wout 59 s |
| QA_optimization_finite_beta.py | (new) | 2810 s | same pattern: last stage 1460 s, Jacobian 1352 s / 5208 columns |

The QH/QI/QP, omnigenity and LASYM scripts were not reached.

### Top bottlenecks across the examples measured
1. **Cold compile.** It was 25-60% of wall in every run. The fixed-boundary single stage spends one 108 s (loaded) XLA compile on the jitted plasma objective, plus 13 s on the refinement factor core, 8 s on the refinement step and about 17 s on hundreds of eager op-by-op compiles. The finite-beta single stage adds virtual casing to that graph: 497 s for the first call at load.
2. **Least-squares Jacobians at high mode number.** One tangent solve per column, so the cost grows with the dof count, and stage max_mode 5 dominates #426's QA.
3. **Movies with field colour.** One equilibrium solve per frame (now off in the fixed single stage).
4. **Free-boundary restart waste,** being fixed in #432.
5. **Virtual casing.** It costs about 107 s in setup (the seed precision plan), about 130 s in the 61x64 final check, and roughly 2-4x on each fixed-boundary trial.
6. **plot_wout.** 23 s CPU cold and 12 s warm; wall at load: gamma_c profile 18 s, 3-D plot_surface 28 s.

### Library speedups, ranked by gain per risk (none implemented; costs measured as above)
1. **Jit `surface_field_values`** (the movie colour) and reuse the optimizer's cached states instead of re-solving each frame. Gain: movie cost from minutes to seconds. Low risk, plotting only.
2. **Make the plot_wout summary cheaper.** Skip the gamma_c profile by default, and use a coarser 3-D surface. Estimated gain: 10-40 s per example. Low risk.
3. **Cut the virtual-casing setup and check.** Profile `plan_vc_precision` and `surface_field_data_from_state`, since the seed plan costs about 100 s under load. A VC grid separate from the coil grid, coarse during the optimization and fine in the check, is an example-level option. It needs an accuracy table first: the vcbench of digits 3 against 4 was interrupted; the first trials agreed to 3e-3 relative in value.
4. **Reduce the plasma-objective compile.** Find what makes the jitted value+grad graph take about 55 s of XLA time (measured at load). Also make sure an optimizer's first accepted step does not recompile (one 28 s compile happened inside the fixed single-stage optimization, after the explicit warm-up). Gain: 20-60 s on every fixed-boundary optimization. Medium risk.
5. **Free boundary: a cheaper anchor** (the Woodbury/NESTOR-coupled preconditioner named in #432) and the restart budget. This is the only route to per-trial cost that fits a 5-minute free-boundary example at ns 31.

### Next steps
1. Run `64d18ed3e`'s finite-beta fixed example end to end, and requalify both fixed examples in a quiet window. Put the measured wall times in the docstrings.
2. Free boundary, both files: run with ns 16 trial solves, MAXITER about 8-10 and MAX_TRIALS about 25, both on base and on #432. Check the saved iterates with the scripts' own checks, and pick the smallest budget that meets every target. Report the boundary and coil change: the owner wants visible optimization, so give max coil displacement and max |dR|, |dZ| against the seed.
3. Augmented-Lagrangian and least-squares companions: the same budget treatment, with the movie off.
4. Finish the QH/QI/QP/LASYM/omnigenity inventory for #426.

rogeriojorge:
## Handoff (paused 2026-09-23, owner out of budget)

**State.** The head is `5f4a5f626`, rebased on main `caaf95301` (#432 is in). The PR is still a draft. CI run 35870689014 on this head was queued behind other PRs and has not reported. Nothing has been merged.

**Budget met on the reference laptop.** Setup: commit `2072cadaa`, which differs from the head only in docstrings and the README. Apple M3 Max, cold JAX cache, one run at a time under a laptop lock. A run was accepted only if it started at 1-minute load <= 10 and its median sampled load was <= 12, including its own run of about 2.5.

| example | wall | median load | targets |
| --- | --- | --- | --- |
| fixed, zero beta | 129 s | 11.1 | all met |
| fixed, 0.5% beta | 199 s | 10.5 | all met |
| free, zero beta | 265 s | 10.9 | all met |
| free, 0.5% beta | 275 s | 8.6 | all met |
| augmented Lagrangian | 154 s | 9.7 | all met |
| least squares | 148 s | 10.1 | all met |

Supporting evidence from office (Xeon, under the shared `flock`, load 5-15): fixed zero beta 333 s, fixed 0.5% beta 622 s, both meeting all targets. So office is about 2.5-3x slower than the laptop. Its free-boundary and companion runs were stopped at the pause.

**To finish:**
1. Wait for CI on `5f4a5f626`, including the fast example tests and the parity lanes.
2. Mark the PR ready, then `gh pr merge 411 -R uwplasma/vmex --squash --admin --delete-branch`.
3. The nightly and weekly lanes run the free-boundary files in smoke mode. Smoke was run on this branch merged with #432 and exits 0; the full nightly lane has not.

**Open levers, not needed for the budget:**
- The #432 follow-up (`perf/free-boundary-anchor-coupling`) cuts the free 0.5% beta file from 397 to 341 s in an A/B at load 10-24, with an identical result.
- Compilation is 35-50% of every run.

rogeriojorge:
Merge coordination: rebased onto main `e205f6962` and stacked in merge order #430 → #441 → #437 → #434 → #436 → #438 → #439 → #411 → #440, so each head carries the PRs before it and its CI tests the tree that lands. Conflicts resolved: CHANGELOG entries kept side by side (the 0.9.1 entry condensed to stay under the 200-line cap), `tests/manifest.json` rows combined, `benchmarks/INDEX.md` regenerated. After each squash-merge the next branch is rebased onto main and merged only when its tree is identical to the tested head.

rogeriojorge:
Paused (owner out of budget) after 0.11.1 was cut from main at `ff97bfa22`; not in 0.11.1. This branch is stacked #438 → #439 → #411 → #440 on main `ff97bfa22`. Its CI run was cancelled to free runners for the release. Next: after #438 merges, `git rebase --onto origin/main <old #438 tip a134e3d82> <this branch>` (for #439), and likewise up the stack; push, wait for the PR gate, squash-merge with `--admin`. Verified on the stack before the pause: both free-boundary single-stage examples meet every target with #432 and #439 in (vacuum: min iota 0.4336, aspect 5.984, B.n/B 0.645 %; 0.5 % beta: min iota 0.4333, aspect 5.947, beta 0.488 %; 8 iterations, 16/17 trials, wall times not meaningful at laptop load ~100).

PR #410, opened 2026-09-21
Require compatible released integration versions

rogeriojorge:
Raise VMEX's sibling dependency floors to the releases that provide the APIs and fixes it uses. This remains a draft until the remaining minimum-version integration gates pass.

| Package | Required floor | Reason |
|---|---|---|
| booz_xform_jax | 0.4.0 | Correct registered-surface selection and cached-grid rebuilding; retains the magnetic-only JAX projection and asymmetric spectra. |
| SOLVAX | 0.21.0 | Accelerator-batched Thomas sweep; later releases add APIs VMEX does not yet require. |
| virtual-casing-jax | 0.0.7, unchanged | Align the compatibility comment and error hint with the existing requirement. |
| ESSOS | 0.17, unchanged | Existing field-line and coil APIs. |
| GKX | 1.8.0 | Adds the `tprim`/`fprim` drive fields used for explicit temperature/density gradients. |

Boozer 0.4.0 is released on PyPI. Its [release notes](https://github.com/uwplasma/booz_xform_jax/releases/tag/v0.4.0) distinguish the host-API fixes from the functional JAX API, which is unchanged from 0.3.0. No new universal transform-speedup claim is made.

### Completed checks

- The actual Boozer 0.4.0 wheel passed four upstream symmetric/asymmetric dense-reference tests, including magnetic-only values and bmnc JVP/VJP: [wheel evidence](https://github.com/uwplasma/vmex/pull/410#issuecomment-5755351706).
- Released Boozer 0.4.0 and SOLVAX 0.21.0, isolated together with Python 3.11.14/JAX 0.9.2 CPU/x64: all **21 Boozer-table, nine plotting, and eight omnigenity tests passed**, including full-marked tests and optimizer composition. Tests explicitly enabled JIT. [Detailed evidence and commands](https://github.com/uwplasma/vmex/pull/410#issuecomment-5755560071).
- Those numerical checks used source `f719c4ff`; the numerical source in this dependency-only branch is unchanged apart from the virtual-casing error hint.
- Updated packaging metadata: **11 passed**. Ruff and whitespace checks pass. The final Boozer-floor update removes five net lines from the dependency comment and adds no files.

The minimum-stack run found a real incompatibility at GKX 1.7.1: explicit
`r_over_lt`/`r_over_ln` failed because its `LinearParams` lacks `tprim` and
`fprim`. The tagged source and released 1.8.0 wheel provide these fields.
Commit `95710eb6` raises the turbulence-extra floor and updates installation
[...]

PR #377, opened 2026-09-19
Let the finite-beta free-boundary example finish and report what it reached

rogeriojorge:
The finite-beta free-boundary example previously raised when its final output solve reached the iteration cap, after optimization had already completed. It also lacked an explicit record of unmet design targets.

This change uses a `[15, NS]` verification ladder, reports capped output solves, writes the same structured summary as the vacuum example, and exits nonzero on unmet targets outside smoke mode. The vacuum example receives the same non-raising verification policy. No library solver code changes.

### Completed full-run evidence

Both runs were measured from clean source `cf9811bf71518a08b1ff315dd4c75f92249e3308` (main `f719c4ff` plus this example change). The existing benchmark record now contains both rows in commit `e53e3b57`. Measurement environment: Linux CPU, Python 3.12.13, JAX/jaxlib 0.11.1, NumPy 2.5.2, SciPy 1.18.1, ESSOS 0.17, four OpenMP threads, persistent compilation cache disabled.

| Full example | Wall | Peak RSS | Trials / optimizer iterations | Final minimum absolute iota | Result |
|---|---:|---:|---:|---:|---|
| Vacuum | 1164.1 s | 4.09 GiB | 33 / 7 | 0.426798 | Output solve converged; stated targets met |
| Finite beta | 1776.2 s | 5.26 GiB | 26 / 6 | 0.258577 | Output solve converged; 0.42 transform floor missed; exit 1 |

Finite-beta final QA is 0.146479, bootstrap residual 0.152869, beta 3.3807%, and aspect 6.9834. L-BFGS-B stopped with `ABNORMAL: `. This is a completed measurement of an unsuccessful design, not a feasibility claim. The older 24.9 GiB finite-beta record predates the Schur memory fix and is superseded for this source.

The timings are diagnostic samples on a shared CPU; load averages and complete provenance are retained. They do not isolate a controlled speedup. Both runs predate #416's failed-trial recovery correction.

At the seed coils, the output ladder converged in 365 iterations with fsqr=9.85e-10 (30.2 s); a single NS=31 rung capped at 2500 iterations with fsqr=5.53e-8 (32.4 s). These seed measurements do not establish the optimized-coil budget comparison.

### Reproduction

At the pinned measurement source:

```sh
OMP_NUM_THREADS=4 VMEX_COMPILATION_CACHE=disabled MPLBACKEND=Agg PYTHONPATH=. \
python benchmarks/single_stage_profile.py --runs free:full free_beta:full \
[...]

PR #413, opened 2026-09-21
Replace the stale plan with current research acceptance gates

rogeriojorge:
Replace the obsolete queue with a current research plan, preserving historical reasoning through immutable links. Main checkpoint: `45f3a7ae`, including #409; VMEX 0.10.0 predates the native-field and surface-call corrections. This documentation-only PR removes roughly 2,000 net lines across existing files.

The next milestone is an integrated tested checkpoint, a controlled collaborator comparison, and an independently verified optimization workflow. Certification, recovery, dependency and profiler corrections come first; factor reuse depends on equilibrium-family and final-physics checks. Winding work, further QI surrogate sweeps and broad coordinate rewrites remain deferred. Explicit maintainer approval is required before any merge; no merge or release is performed here.

Current evidence and remaining gates:

- #417/#418 anchor derivatives to measured states and enforce direct admission. #421 aligns callback devices; #422 adds complementary placement tests. #423 measures the actual supplied coefficients instead of silently replacing the edge. Follow-up `02bf33d9` resolves the reviewed rounding and public-helper device-placement risks: eight focused cases pass with all nine new executable source statements covered; the combined penalty/device suite passes all 71 tests on JAX 0.10.1/SOLVAX 0.21.0 with two CPU devices. An additional [explicit CPU/CUDA callback check](https://github.com/uwplasma/vmex/pull/423#issuecomment-5757198007) passes on NS5/MPOL3 Solovev, with coefficient differences below 3.2e-15. Asymmetric accelerator cases and full optimization remain open.
- #416 restores deterministic cold recovery. One pressure observable reaches about 0.017% adjoint/re-solve agreement. For the actual finite-beta objective, the exact-reference directional gap falls from 2.1746% at h=1e-4 to 0.08114% at h=5e-5, with both pairs admitted and no cold recovery. This improves local agreement but does not establish an asymptotic window, bound root error, or qualify the final design. [Latest controlled record](https://github.com/uwplasma/vmex/pull/377#issuecomment-5757012060). The existing finite-beta optimized design still misses its transform floor.
- #419 certifies reused response columns, but independent QI reconvergence changes the frozen m=1 residual family and fine-grid mirror checks fail. The canonical post-solve experiment also fails root admission; an accurate first linear solve produces a catastrophic full Newton step. The [bounded globalized follow-up](https://github.com/uwplasma/vmex/pull/419#issuecomment-5757238643) takes eight admissible-geometry steps but remains far from root admission. Stop extending post-solve recovery and inspect initialization-time pinning before promoting reuse. [Public reproducer](https://github.com/uwplasma/vmex/pull/419#issuecomment-5756859333).
- The scalar factor cache and bounded QI surrogate were rejected on complete-solve cost and independently checked design quality. Do not revive them on kernel timing or a local Taylor test alone.
- The 3-D QA dense reference reduces independent force by 25% in one admissible step, but does not achieve the 10x or nonlinear-stationarity gates. Its 604-second, 11.13-GiB diagnostic is not a production solver recommendation. [Source, controls and results](https://github.com/uwplasma/vmex/pull/413#issuecomment-5756509703). SOLVAX #119 supplies diagnostic information; recursive residuals are not independent certificates.
- #410 qualifies proposed released dependency floors with targeted integration checks. #420 corrects warm profiling to repeat every workflow stage; affected historical aggregate timings require remeasurement. The latest main CI run and recent preceding runs were cancelled, so there is no fresh integrated green checkpoint.

The collaborator's complete VMEX 0.3 versus available 0.7/0.8/current comparison remains unresolved. Use identical objectives, accuracy, constraints and final fine-grid acceptance, with separate cold/cache-reload/warm timing and actual device placement. Broaden to QA/QH/QI/QP and single-stage CPU/GPU workflows only after the first controlled cases pass. Keep plotting diagnostics separate from solving and rendering when profiling.

Documentation changes also correct the adjoint equations and qualify GPU placement as a workload heuristic. The generated capability table now distinguishes discrete-equation derivatives and selected observable checks from full design qualification. The README now uses the compact WOUT export helper and removes an unqualified exterior-field showcase. Exterior documentation and the example docstring now match the implementation warning about near-surface continuation; its supplied-field variant still needs independent qualification. Generic algorithms belong in SOLVAX; VMEX owns physical operators and invalidation.

Validation: this PR changes documentation only. Whitespace and exact outgoing text were reviewed, including privacy. The new #423 checks above are recorded at their tested revision; prior documentation builds and other numerical checks retain only their original scope. No merge or release is included.

Latest documentation validation: the generated capability page matches its canonical data; all four capability guards pass, along with prose/media and whitespace checks. The [complete base-linearization capture](https://github.com/uwplasma/vmex/pull/377#issuecomment-5757443492) now reproduces the old base exactly and certifies the repeated adjoint to a true relative linear residual of 2.87e-11. The empirical endpoint chord has a 2.31% tangent defect relative to its field contribution and changes a frozen m=1 component. Next separate that complement contribution and replay existing endpoints against the same base operator, without new endpoint solves. The public record is explicit that exact replay still requires publication of the tapes or requalification of regenerated public inputs. #407 is an unfinished CI-capacity handoff, not an implemented lane split.

Documentation follow-up `f17cf6d9`: corrected #422’s status (merged into the #421 feature branch, not main). Prose/media and whitespace checks pass; the tracing example’s executable AST is unchanged. The README is eight lines shorter, no files or numerical behavior were added, and the underlying illustration is retained.


rogeriojorge:
### Bounded QI smoothness evidence and public reproduction

Baseline: [`f719c4ff503c526368254aebf3ba9a9119c17f87`](https://github.com/uwplasma/vmex/tree/f719c4ff503c526368254aebf3ba9a9119c17f87). This probes the public frozen Boozer spectrum in `tests/data/boozer_nfp4_QI_finite_beta.npz`, without solving an equilibrium or running an optimizer. Perturb the outer-surface cosine coefficient `(m,n)=(0,4)`. A sampled well-minimum cell transition occurs near `t=8.87608289028498e-4`.

The preserved baseline experiment recorded:

| step h | positive Taylor remainder | negative Taylor remainder | centered derivative error |
|---|---:|---:|---:|
| 1e-5 | 1.086e-13 | 6.928e-7 | 3.464e-2 |
| 1e-6 | 1.085e-15 | 6.928e-7 | 3.464e-1 |
| 1e-7 | 1.057e-17 | 6.928e-7 | 3.464e0 |
| 1e-8 | 7.861e-19 | 6.928e-7 | 3.464e1 |

The baseline value was `0.001080608989122136` and its AD derivative `-0.0002960530339699146`. The nonvanishing one-sided remainder means this sampled switch does not support a smooth-gradient interpretation. Away from the transition the reported remainders were quadratic. This is a local objective diagnostic, not evidence about equilibrium derivative accuracy or complete optimization runtime.

Run the following from a checkout of the pinned revision with VMEX dependencies installed. Enable x64 as shown. The historical environment was not captured in this compact record, so the rounded measurements are observations rather than a cross-platform golden tolerance; independently check the transition if rounding moves it.

```sh
python - <<'PYCODE'
"""Reproduce the bounded constructed-QI well-switch experiment.

Run from the pinned VMEX checkout.
"""

from __future__ import annotations

import jax
import jax.numpy as jnp
import numpy as np

from vmex.core import optimize as opt

jax.config.update("jax_enable_x64", True)
print("jax", jax.__version__, "numpy", np.__version__, "backend", jax.default_backend())

QI_OPTIONS = {
    "nphi": 61,
    "nalpha": 13,
    "n_bounce": 21,
    "include_bounce_endpoints": True,
    "softness": 2.0e-2,
    "width_weight": 1.0,
    "branch_width_weight": 0.5,
    "branch_width_softness": 2.0e-2,
    "profile_weight": 0.1,
    "shuffle_profile_weight": 1.0,
    "shuffle_profile_softness": 2.0e-2,
}
T_SWITCH = 8.87608289028498e-4
SURFACE_INDEX = 2
MODE_INDEX = 1
STEPS = (2e-4, 1e-4, 5e-5, 2e-5, 1e-5, 5e-6, 2e-6,
         1e-6, 5e-7, 2e-7, 1e-7, 5e-8, 2e-8, 1e-8)


def main() -> None:
    with np.load("tests/data/boozer_nfp4_QI_finite_beta.npz") as data:
        bmnc = jnp.asarray(data["bmnc_b"].T)
        xm = jnp.asarray(data["ixm_b"])
        xn = jnp.asarray(data["ixn_b"])
        iota = jnp.asarray(data["iota_b"])
        nfp = int(data["nfp_b"])

    assert (int(xm[MODE_INDEX]), int(xn[MODE_INDEX])) == (0, 4)

    def objective(t):
        spectrum = bmnc.at[SURFACE_INDEX, MODE_INDEX].add(t)
        return opt.quasi_isodynamic_residual(
            bmnc_b=spectrum,
            xm_b=xm,
            xn_b=xn,
            iota_b=iota,
            nfp=nfp,
            **QI_OPTIONS,
        )["total"]

    value_and_grad = jax.jit(jax.value_and_grad(objective))
    value_and_grad(jnp.asarray(T_SWITCH))[0].block_until_ready()
    center, gradient = map(float, value_and_grad(T_SWITCH))
    base, base_gradient = map(float, value_and_grad(0.0))
    print(f"mode=(0,4) surface={SURFACE_INDEX} t_switch={T_SWITCH:.15g}")
    print(f"base={base:.16g} base_gradient={base_gradient:.16g}")
    print(f"center={center:.16g} gradient={gradient:.16g}")
    for step in STEPS:
        minus, gradient_minus = map(float, value_and_grad(T_SWITCH - step))
        plus, gradient_plus = map(float, value_and_grad(T_SWITCH + step))
        remainder_plus = abs(plus - center - step * gradient)
        remainder_minus = abs(minus - center + step * gradient)
        centered_error = abs((plus - minus) / (2.0 * step) - gradient)
        print(
            f"h={step:.1e} fm={minus:.16g} fp={plus:.16g} "
            f"gm={gradient_minus:.9g} gp={gradient_plus:.9g} "
            f"rp={remainder_plus:.3e} rm={remainder_minus:.3e} "
            f"centered_error={centered_error:.3e}"
        )


if __name__ == "__main__":
    main()

PYCODE
```

Next gate: compare one subcell alternative using the same boundary, constraints and budgets, then validate the final design on independent fine Boozer grids and transport diagnostics. No production surrogate or physical design-quality claim is established by this probe.

rogeriojorge:
### QI smoothness: paired solved-design checkpoint

The frozen-spectrum Taylor probe above establishes local differentiability, but the acceptance criterion is the quality of a design produced from the same seed and constraints. I therefore ran the unchanged full QI example in two separate clean checkouts:

- current objective: `f719c4ff503c526368254aebf3ba9a9119c17f87`;
- smooth-well candidate: `413d7fd2c59797ede19f09ce1c5319d6d31577cf` (`qi/smoothness-bounded`);
- identical `benchmarks/qi_optimization_profile.py` in both checkouts (SHA-256 `a1de14694bb3ffacc1f74ac12fcabaea1fe24646059b4d91bb20f727fb3fd480`);
- Python 3.12.13, JAX/JAXlib 0.11.1, booz-xform-jax 0.4.0, SOLVAX 0.24.0, NumPy 2.5.2, SciPy 1.18.1, ESSOS 0.17;
- CPU-only execution with `OMP_NUM_THREADS=4`, `OPENBLAS_NUM_THREADS=4`, and `MKL_NUM_THREADS=4`.

| Quantity | Current objective | Smooth-well candidate |
|---|---:|---:|
| constructed QI | 0.003027560 | 0.004690223 |
| independently refined-grid QI | 0.002767 | 0.004558 |
| least-squares cost | 0.0311670 | 0.0354837 |
| reported optimality | 0.2393 | 0.6345 |
| `nfev` / `njev` | 20 / 11 | 20 / 14 |
| rejected trials | 0 | 1 |
| aspect ratio | 7.5333 | 7.2025 |
| minimum `|iota|` | 0.539276 | 0.535605 |
| mirror ratio | 0.2088 | 0.2089 |
| elongation | 7.3050 | 6.7266 |
| all reported constraints met | yes | yes |
| wall time / peak RSS | 295 s / 3.06 GiB | 205 s / 3.26 GiB |

The candidate remained feasible, but its constructed-QI value was 54.9% higher and its refined-grid value was 64.7% higher. The wall times were collected under different shared-machine load and compilation state, so they are recorded for completeness rather than treated as a performance comparison. This candidate therefore does not meet the physical-design gate and should not become a production PR in its present form.

Reproduction uses the same command independently in each pinned checkout; `--baseline-ref` is deliberately avoided because it changes the example source while retaining the importing checkout's objective implementation:

```sh
export JAX_PLATFORMS=cpu JAX_ENABLE_X64=True XLA_PYTHON_CLIENT_PREALLOCATE=false
export OMP_NUM_THREADS=4 OPENBLAS_NUM_THREADS=4 MKL_NUM_THREADS=4
python benchmarks/qi_optimization_profile.py \
  --runs full --timeout 1800 --output result.json
```

The next bounded diagnostic is saved-endpoint cross-evaluation under both objective definitions, decomposed into occupancy width, branch width, profile, and shuffle-profile contributions. No optimizer restart or surrogate-width variation is justified before that attribution is available.

rogeriojorge:
### Saved-boundary cross-evaluation

I re-solved each profiler-captured final input at its saved `ns=101`, `ftol=1e-14`, `niter=8000` settings, generated each coarse and fine Boozer spectrum once, and passed that identical frozen spectrum through both objective definitions. This prevents equilibrium or transform differences from contaminating the definition comparison.

Both cold re-solves reached the requested force tolerance. The current-objective boundary finished at `FSQR=9.89e-15`; the smooth-objective boundary finished at `FSQR=9.93e-15`.

| Saved boundary | Grid | Current definition | Smooth definition | Smooth - current |
|---|---|---:|---:|---:|
| current-objective result | `61 x 18 x 21` | 0.003029036 | 0.003044486 | +0.510% |
| current-objective result | `101 x 29 x 31` | 0.002765113 | 0.002769066 | +0.143% |
| smooth-objective result | `61 x 18 x 21` | 0.004560165 | 0.004566641 | +0.142% |
| smooth-objective result | `101 x 29 x 31` | 0.004452189 | 0.004447255 | -0.111% |

Thus, under the same definition, the smooth run's boundary is worse by 50.5% on the optimization grid and 61.0% on the refined grid using the current definition; the corresponding differences using the smooth definition are 50.0% and 60.6%. The direct definition change at either fixed boundary is at most 0.51%, so metric distortion does not explain the solved-design gap.

Decomposing the squared residual under the current definition gives the following increase from the current run's boundary to the smooth run's boundary:

| Residual family | Optimization grid | Refined grid |
|---|---:|---:|
| occupancy width | +81.6% | +78.8% |
| branch width | +139.8% | +150.9% |
| profile consistency | +28.5% | +28.6% |
| shuffle profile | +31.1% | +46.4% |

All four physical residual families degrade; branch width degrades most. The evidence points to the small surrogate change altering the optimization trajectory and reaching an inferior basin, rather than merely changing how the same endpoint is scored.

Reproducer details:

1. Use current objective commit `f719c4ff503c526368254aebf3ba9a9119c17f87` and smooth candidate `413d7fd2c59797ede19f09ce1c5319d6d31577cf` with the dependency pins in the preceding comment.
2. Cold-solve each profiler-captured final input using its embedded radial resolution and tolerances.
3. For each converged state, call `boozer_modes_from_wout` once with `(mboz, nboz)=(12,12)` and once with `(14,14)` on `surfaces=np.linspace(0.1, 1.0, 6)`.
4. Feed each frozen spectrum to each commit's `quasi_isodynamic_residual`, using `(nphi,nalpha,n_bounce)=(61,18,21)` and `(101,29,31)` respectively.
5. Split `residuals1d` in its documented order and sum squares for occupancy width, branch width, profile consistency, and shuffle profile.

One limitation is that the profiler retained the final boundary inputs rather than the warm-started final equilibrium states. The cold current-boundary score reproduces its earlier refined score within 0.1%, while the cold smooth-boundary score is about 2.4% lower than its earlier refined score. This does not affect the cross-definition ranking above, but exact warm-state reproduction would require retaining the final `wout` in a future profiler run.

This closes the attribution gate for the present candidate: do not open a production PR or vary the surrogate width yet. A future alternative should preserve the current boundary's physical score while addressing branch-selection differentiability, and should retain the final state for exact cross-evaluation.

rogeriojorge:
Resolved analytic-recovery gate (CPU/x64, source `f719c4ff`, JAX/JAXLIB 0.11.1)

This is a bounded local-recovery test, not evidence that the 3D force plateau is upstream of the nonlinear solver. I projected the analytic finite-beta Solov'ev state from `tests/test_strong_force_solovev.py` into the native degree-5, four-span, `mmax=8` basis. I then added a 0.1% projected interior perturbation in one `R_cos(m=2)` and one `L_sin(m=1)` direction and ran the production rectangular Gauss-Newton polish with the same fixed profiles and structured chart.

| independent/off-grid measure | analytic reference | perturbed | 12 accepted steps | 40 accepted steps |
|---|---:|---:|---:|---:|
| pointwise normalized force L2 | 2.4545e-6 | 4.1708e-2 | 3.0149e-5 | 2.3370e-5 |
| dimensional force L2 (N/m^3) | 4.1359e-1 | 2.1547e3 | 3.9589 | 2.5889 |
| collocation RMS | 1.4432e-6 | 6.0419e-3 | 1.2019e-5 | 8.2030e-6 |
| relative off-grid B difference from reference | 0 | 3.9218e-5 | 1.4688e-6 | 1.2799e-6 |
| relative off-grid J difference from reference | 0 | 2.1541e-3 | 4.9911e-5 | 6.3858e-5 |
| least-squares relative optimality | -- | -- | 5.3999e-5 | 2.4983e-5 |

Both solves accepted every nonlinear step and preserved `phipf`, `chipf`, and pressure exactly; boundary and gauge residuals remained zero and the 40-step minimum signed Jacobian was 0.89996. The 40-step solve improved normalized force by 1,785x and dimensional force by 832x from the perturbation. It nevertheless remained 9.52x and 6.26x above the corresponding analytic-reference values. It also did **not** satisfy least-squares stationarity (`least_squares_success=false`), despite all 40 steps being accepted and damping reaching 1e-12. The off-grid B error improved slightly from 12 to 40 steps, while the off-grid J error worsened from 4.9911e-5 to 6.3858e-5. This rules out a simple insufficient-iteration-budget explanation and identifies a solver/chart/scaling plateau even for this resolved axisymmetric analytic case.

Compact reproduction recipe using existing public helpers:

1. Construct `_analytic_state(_Solovev(mmax=8), degree=5, spans=4, mmax=8)` from `tests/test_strong_force_solovev.py`.
2. Build the low-order preconditioner from `examples/data/input.solovev` at `mpol=9`, `ntor=0`, `ntheta=24`, `nzeta=4`, `ns=19`, `ftol=1e-10`, and `niter=1000`.
3. Create a projected high-order correction with `R_cos[2,1] = 1e-3 * ||R_cos[2]||` and `L_sin[1,1] = -1e-3 * ||L_sin[1]||`; build the structured chart with one balance iteration and two scale probes.
4. Call `polish_collocation_least_squares` with tolerance 1e-8, validation tolerance 1e-4, radial-refinement tolerance 1e-1, initial damping 1e-3, Krylov restart 20x10, and either 12 or 40 nonlinear steps.
5. Independently evaluate `certify_strong_force(..., angular_multiplier=1)` and off-grid position/B/J at the Cartesian product of `rho=[0.17,0.39,0.68,0.91]`, `theta=[0.23,1.19,2.71,4.83]`, with `zeta=linspace(0.11,5.71,16)`.

The next useful discriminator is one fixed-endpoint linear-algebra diagnostic at the 40-step state: compute the scaled residual Jacobian's smallest right-singular directions and project both the gradient and the 12-to-40 displacement onto them. That separates a near-null/scaling-limited plateau from damping or inaccurate Krylov steps without spending another nonlinear iteration budget.

rogeriojorge:
Fixed-endpoint linear-solver discriminator (same analytic case and source pin as the preceding comment)

The deterministic reconstruction matched the prior 40-step endpoint: normalized force L2 was 2.33604e-5 versus 2.33697e-5 previously (0.04% difference), while the least-squares cost and gradient matched exactly. The endpoint was checkpointed before constructing the 1184-by-144 Jacobian.

At the endpoint, the scaled collocation Jacobian has full numerical rank 144 and condition number 1.12e6. The two-probe column-norm estimate is imperfect but not grossly wrong: estimated/exact ratios are 0.163 minimum, 0.832 median, and 2.343 maximum. Damping is already at 1e-12.

| one-step result at identical damping | dense augmented least squares | production PCG (200 steps) |
|---|---:|---:|
| normal-equation relative residual | 8.19e-9 | 2.30e-1 |
| actual/predicted reduction | 0.9999999 | 1.0000003 |
| scaled collocation residual L2 | 2.069e-4 | 4.644e-2 |
| independent normalized force L2 | 6.618e-8 | 2.335e-5 |
| dimensional force L2 (N/m^3) | 1.252e-2 | 2.564 |
| off-grid B difference from analytic reference | 3.618e-9 | 1.276e-6 |
| off-grid J difference from analytic reference | 1.577e-7 | 6.415e-5 |

The endpoint before either trial had scaled collocation residual 4.672e-2, normalized force L2 2.336e-5, dimensional force L2 2.584 N/m^3, off-grid B difference 1.280e-6, and off-grid J difference 6.386e-5. Both trial steps preserved the fixed profiles and physical Jacobian margin; the dense trial's minimum signed Jacobian was 0.90000. The PCG solve exhausted all 200 iterations, did not converge, and its direction differed from the dense direction by 0.99996 in relative L2.

This isolates the plateau in this resolved analytic case to the unpreconditioned Krylov solve of the normal equations. It does not establish the cause of the separate 3D representation plateau.

Compact reproduction using public helpers: reconstruct the 40-step endpoint by the preceding recipe; define `r(x) = strong_collocation_residual(variable_scale*x, runtime, chart) / collocation_scale`; form `J = jax.jacfwd(r)(x40)`; compare `numpy.linalg.lstsq(vstack([J, sqrt(mu)*I]), concatenate([-r(x40), zeros(144)]))` against `solvax.pcg` applied to `(J.T@J + mu*I)` with `mu=1e-12`, `rtol=1e-3`, and 200 steps; then pass each trial state through `certify_strong_force(..., angular_multiplier=1)` and the same off-grid B/J nodes.

The smallest justified implementation experiment is now a normal-equation preconditioner test at this saved endpoint, with the same one-step linear residual and independent physical checks. Increasing nonlinear iterations or retuning damping is not supported by these data.

rogeriojorge:
Saved-endpoint policy/preconditioner follow-up

I reused the checkpointed 40-step endpoint from the preceding comparison; no nonlinear reconstruction or iteration-budget sweep was performed.

| bounded check | result |
|---|---|
| require every PCG solve to meet `rtol=1e-3` | 12/12 trials rejected; 2,400/2,400 PCG iterations; damping rose from 1e-12 to 1.68e-5; force and observables were unchanged |
| rank-64 public `solvax.nystrom_preconditioner`, one GN step | trust-ratio accepted; scaled collocation L2 4.672e-2 -> 2.814e-2; normalized force L2 2.336e-5 -> 1.764e-5; dimensional force L2 2.584 -> 1.640 N/m^3 |

The preconditioned trial improved the off-grid B difference from 1.280e-6 to 8.171e-7, while the off-grid J difference worsened from 6.386e-5 to 7.372e-5. It exhausted all 200 PCG steps and did not meet the configured tolerance. The step was accepted by the inexact Gauss-Newton trust ratio; it was not independently certified as a stationary solver root.

For a separate same-endpoint preconditioned PCG direction, a fresh application of the normal operator gave relative residual 0.21973430839 versus PCG's recursively reported 0.21973430833. This is only marginally below the unpreconditioned fixed-endpoint value 0.22976047658. The GN-wrapped accepted direction reported 0.4088; that exact direction was not persisted, so the fresh check must not be attributed to it.

These results reject strict inner convergence as a recovery policy for this endpoint and reject the tested rank-64 Nyström callback as the conditioning fix. They support exposing inner-solve status while preserving useful inexact-step acceptance by default. The focused SOLVAX diagnostics change is [uwplasma/SOLVAX#119](https://github.com/uwplasma/SOLVAX/pull/119). No 3D source/representation conclusion follows from this axisymmetric case.

rogeriojorge:
Saved-endpoint original-operator least-squares comparison

I re-materialized the 1184-by-144 scaled Jacobian and residual at the checkpointed 40-step endpoint, then compared SciPy 1.18.1 LSQR and LSMR at the same 200-iteration budget. Each method used exactly 200 `J` and 201 `J.T` products with damping `sqrt(1e-12)`. The true normal residual below was recomputed from a fresh dense `J.T @ (J @ step + r) + mu * step`, rather than taken from a recursive Krylov estimate.

| method | true normal relative residual | relative direction error vs dense | independent force L2 | off-grid J difference |
|---|---:|---:|---:|---:|
| dense augmented least squares | 4.64e-9 | 0 | 6.618e-8 | 1.577e-7 |
| production PCG reference | 2.30e-1 | 9.9996e-1 | 2.335e-5 | 6.415e-5 |
| SciPy LSQR, 200 iterations | 2.486 | 9.9985e-1 | 2.342e-5 | 6.584e-5 |
| SciPy LSMR, 200 iterations | 2.194e-2 | 9.9998e-1 | 2.354e-5 | 6.396e-5 |

Both SciPy solvers stopped at their iteration cap. LSQR is worse than PCG on the true normal residual and does not justify a larger run. LSMR reduces that residual by 10.5x relative to PCG, but its 200-iteration direction remains far from dense and does not improve the independent physical result. One bounded LSMR continuation to 2,000 iterations on the persisted matrix/residual is the next gate; no new solver implementation or production performance claim is warranted from the 200-iteration result.

rogeriojorge:
The bounded LSMR continuation also stopped at its cap: 2,000 iterations, 2,000 `J` products, and 2,001 `J.T` products. A fresh normal-operator application gave relative residual 2.416e-3, still 2.42x above the 1e-3 linear gate. The direction remained far from the dense reference (relative difference 0.9873; norm 50.9 versus 741.1 for dense).

The true normal residual continued to improve from 2.194e-2 at 200 iterations, but the tenfold larger operator budget neither met the configured tolerance nor materially approached the dense step. Because the linear gate failed, I did not spend another equilibrium-runtime or physical-certificate evaluation. This closes the prototype without a SOLVAX LSQR/LSMR implementation; the evidence still points to conditioning/preconditioning as the unresolved numerical issue.

rogeriojorge:
### Bounded production-QA dense linear discriminator

This closes the proposed dense fallback check at the **initial** production QA polish point. It does not reopen E2's existing 3-D reachability verdict and does not promote a dense production solver.

The frozen operator was the exact production collocation residual in the production structured chart,

```python
r(y) = strong_collocation_residual(variable_scale * y, runtime, chart) / residual_scale
```

with the production initial damping `lambda = 1e-3`. Thus the augmented direct problem used `sqrt(lambda) I` in the same scaled `y` coordinates as production Gauss--Newton. The run admitted only the expected `18018 x 1336` Jacobian, required positive finite damping, compared independent LAPACK GELSD/GELSY directions, recomputed the original damped-normal residual, checked a deterministic symmetry/positive-Rayleigh probe, and ran SOLVAX PCG on the **stored-J** normal operator at production `linear_rtol=1e-3`, `linear_max_steps=600`. Independent force evaluation was gated on the direct linear checks.

| quantity | result |
|---|---:|
| GELSD / GELSY rank | 1336 / 1336 |
| GELSD / GELSY fresh original-normal residual | `8.39e-15` / `4.72e-15` |
| GELSD--GELSY direction difference | `3.41e-14` |
| normal-operator probe asymmetry | `4.55e-16` |
| stored-J PCG | converged, 324 / 600 iterations |
| PCG reported / fresh original-normal residual | `8.6446e-4` / `8.6446e-4` |
| PCG--GELSD direction difference | `5.23e-2` |
| initial / dense linearized augmented cost | `9009.0` / `4048.73` |
| PCG linearized augmented cost | `4049.64` |
| accepted direct fraction / trust ratio | `1.0` / `1.00007` |
| independent absolute L2 | `34273.62 -> 25697.88` (1.334x) |
| independent `eps_F` | `0.377400 -> 0.330834` (1.141x) |
| near-axis / bulk / edge absolute L2 | `97262 -> 78942` / `21864 -> 20842` / `35398 -> 17391` |
| minimum signed Jacobian | `0.0100157 -> 0.0100234` |
| setup / explicit J / both factorizations | `203.04 s` / `297.98 s` / `15.05 s` |
| total / peak RSS | `603.75 s` / `11.13 GiB` |

The direct linear solve is excellent, but the configured PCG tolerance is already met at this point and its linearized cost is essentially the dense reference. The direct step improves the independent physical measures modestly, not by E2's 10x gate. Dense integration is therefore not justified here: explicit-J assembly costs about 20 times the two direct factorizations and the run peaks above 11 GiB. This does **not** establish matrix-free/stored-J bit parity, stationarity after one step, a general 3-D limitation, or the cause of later low-damping PCG exhaustion. A later fixed endpoint with demonstrated inner starvation would be needed before reconsidering a bounded dense option.

Reproduction pins: VMEX `45f3a7aeaeedd50ee7a04d553bc0d281bb8510a4`, Python 3.12.13, JAX 0.9.2, SOLVAX 0.20.0, NumPy 2.5.2, SciPy 1.18.1, float64 CPU. The process was confined to 16 CPU cores; BLAS/OpenMP thread counts were one; the script had a 1500 s alarm inside an 1800 s process timeout and an external 24 GiB measured-RSS kill gate. The checkpoint checksums were `29d80e99...bf54e2b` for J/r/scales/damping and `659c22aa...75cff` for the three directions; arrays are not attached.

```bash
taskset -c 0-15 env \
  JAX_PLATFORMS=cpu JAX_ENABLE_X64=1 CUDA_VISIBLE_DEVICES= \
  OMP_NUM_THREADS=1 OPENBLAS_NUM_THREADS=1 MKL_NUM_THREADS=1 \
  BLIS_NUM_THREADS=1 NUMEXPR_NUM_THREADS=1 \
  python benchmarks/e2_dense_reference.py \
  --case qa-production --steps 1 --damping 1e-3 \
  --jacobian-chunk-size 8 --memory-budget-gib 2 \
  --budget-seconds 1500 --linear-data results/linear.npz \
  --output results/report.json
```

<details>
<summary>Exact uncommitted benchmark-only patch used for the measurement</summary>

Base SHA: `45f3a7aeaeedd50ee7a04d553bc0d281bb8510a4`  
Patch SHA-256: `bfb0eca0751a9f489e727be1b0ef540bf9f6fe6789820c37e72616410c0de285`  
Resulting source SHA-256: `dff956624ead8870f12f97c75b08909c3a0bea08af12e8c72e5be34849c48e9d`

```diff
diff --git a/benchmarks/e2_dense_reference.py b/benchmarks/e2_dense_reference.py
index b77ad6fd..7ba982f1 100644
--- a/benchmarks/e2_dense_reference.py
+++ b/benchmarks/e2_dense_reference.py
@@ -11,6 +11,7 @@ https://docs.scipy.org/doc/scipy/reference/generated/scipy.linalg.lstsq.html
 from __future__ import annotations
 
 import argparse
+import gc
 from importlib.metadata import version
 import json
 from pathlib import Path
@@ -29,12 +30,35 @@ import numpy as np
 from scipy.linalg import lstsq
 
 import vmex
+from solvax import chunked_jacfwd, pcg
 from benchmarks._provenance import assert_repo_vmex, git_state
 from benchmarks.e1_functional_consistency import qa_state, shaped_state
-from vmex.core.polish import _flatten_high, _physical_equation_basis, apply_high_order_correction
+from vmex.core import implicit
+from vmex.core.input import VmecInput
+from vmex.core.multigrid import solve_multigrid
+from vmex.core.polish import (
+    _flatten_high,
+    _physical_equation_basis,
+    apply_high_order_correction,
+    build_low_order_preconditioner,
+    make_strong_root_runtime,
+    make_strong_structured_chart,
+    strong_collocation_residual,
+)
+from vmex.core.polish_driver import (
+    PolishConfig,
+    _collocation_variable_scale,
+    _corrected_state,
+)
 from vmex.core.profiles import MU0
 from vmex.core.radial_basis import _span_quadrature
-from vmex.core.strong_force import _basic_fields, _point_force, certify_strong_force
+from vmex.core.strong_force import (
+    _basic_fields,
+    _point_force,
+    certify_strong_force,
+    evaluate_strong_force,
+    lift_high_order_state,
+)
 
 
 def certificate(state):
@@ -51,6 +75,316 @@ def spectrum(matrix):
             "singular_values": singular.tolist()}
 
 
+def _production_dense_bytes(rows: int, columns: int) -> dict[str, int]:
+    """Conservative host-array estimate, excluding JAX/XLA intermediates."""
+
+    itemsize = np.dtype(np.float64).itemsize
+    jacobian = rows * columns * itemsize
+    augmented = (rows + columns) * columns * itemsize
+    normal_workspace = columns * columns * itemsize
+    # J remains live while an augmented matrix and one LAPACK-owned factor
+    # copy are present. Eight n-by-n arrays allow for divide-and-conquer SVD
+    # workspace and cross-check directions without pretending to bound the
+    # separate JAX/XLA differentiation footprint.
+    admitted_peak = 2 * jacobian + 2 * augmented + 8 * normal_workspace
+    return {
+        "jacobian": jacobian,
+        "augmented_matrix": augmented,
+        "normal_matrix": normal_workspace,
+        "admitted_host_peak": admitted_peak,
+    }
+
+
+def _production_qa_context():
+    """Build the exact default production QA polish chart at its seed state."""
+
+    inp = VmecInput.from_file(ROOT / "examples/data/input.nfp2_QA_smooth_beta")
+    result = solve_multigrid(inp, polish_force_balance=False)
+    config = PolishConfig(fail_policy="return_unpolished")
+    ns = int(np.asarray(inp.ns_array)[-1])
+    implicit_config = implicit.make_config(
+        inp, ns=ns, lconm1=True, multigrid=False
+    )
+    params = implicit.params_from_input(inp)
+    legacy_runtime = implicit.runtime_from_params(params, implicit_config)
+    dof_mask = implicit._dof_mask(result.state, legacy_runtime, implicit_config)
+    refined_state = implicit._refined_state(
+        implicit_config, params, result.state, dof_mask
+    )
+    native = lift_high_order_state(
+        refined_state, legacy_runtime, degree=config.radial_degree
+    )
+    low_preconditioner = build_low_order_preconditioner(
+        native,
+        params,
+        implicit_config,
+        refined_state,
+        dof_mask,
+        probe_chunk_size=4,
+    )
+    runtime = make_strong_root_runtime(
+        native,
+        low_preconditioner,
+        dof_mask,
+        balance_full_root=False,
+        radial_quadrature_order=config.radial_quadrature_order,
+    )
+    chart = make_strong_structured_chart(runtime)
+    zero = jnp.zeros((int(chart.size),), dtype=jnp.asarray(native.R_cos).dtype)
+    initial = strong_collocation_residual(zero, runtime, chart)
+    residual_scale = jnp.asarray(max(
+        float(jnp.linalg.norm(initial)) / np.sqrt(float(initial.size)),
+        1.0e-12,
+    ))
+    variable_scale = jnp.asarray(_collocation_variable_scale(
+        runtime,
+        chart,
+        residual_scale,
+        zero,
+        int(initial.size),
+        config.collocation_scale_probes,
+    ))
+    return runtime, chart, residual_scale, variable_scale
+
+
+def production_qa_discriminator(
+    damping: float,
+    memory_budget_gib: float,
+    jacobian_chunk_size: int,
+    linear_data: Path | None,
+):
+    """One admitted dense direction for the production QA polish objective."""
+
+    setup_started = time.perf_counter()
+    print("building exact production QA chart", flush=True)
+    runtime, chart, residual_scale, variable_scale = _production_qa_context()
+    setup_seconds = time.perf_counter() - setup_started
+    zero = jnp.zeros((int(chart.size),), dtype=variable_scale.dtype)
+
+    def residual(vector):
+        return strong_collocation_residual(
+            variable_scale * vector, runtime, chart
+        ) / residual_scale
+
+    r = np.asarray(residual(zero))
+    rows, columns = int(r.size), int(zero.size)
+    if (rows, columns) != (18018, 1336):
+        raise RuntimeError(
+            f"production QA shape changed from 18018x1336 to {rows}x{columns}"
+        )
+    memory = _production_dense_bytes(rows, columns)
+    memory_budget = int(memory_budget_gib * 1024**3)
+    if memory["admitted_host_peak"] > memory_budget:
+        raise MemoryError(
+            "admitted dense host arrays exceed the configured memory budget"
+        )
+
+    print(
+        f"assembling admitted {rows}x{columns} Jacobian in "
+        f"chunks of {jacobian_chunk_size}",
+        flush=True,
+    )
+    jacobian_started = time.perf_counter()
+    jacobian = np.asarray(chunked_jacfwd(
+        residual, chunk_size=jacobian_chunk_size
+    )(zero))
+    jacobian_seconds = time.perf_counter() - jacobian_started
+    gradient = jacobian.T @ r
+    rhs = np.concatenate((-r, np.zeros(columns)))
+    augmented = np.vstack((jacobian, np.sqrt(damping) * np.eye(columns)))
+    if linear_data is not None:
+        np.savez(
+            linear_data,
+            jacobian=jacobian,
+            residual=r,
+            variable_scale=np.asarray(variable_scale),
+            residual_scale=np.asarray(residual_scale),
+            damping=np.asarray(damping),
+        )
+        print("saved fixed-endpoint Jacobian and residual", flush=True)
+
+    print("factoring augmented system with GELSD and GELSY", flush=True)
+    solve_started = time.perf_counter()
+    svd_direction, _, rank, _ = lstsq(
+        np.array(augmented, order="F", copy=True),
+        rhs,
+        lapack_driver="gelsd",
+        overwrite_a=True,
+        overwrite_b=False,
+    )
+    gc.collect()
+    qr_direction, _, qr_rank, _ = lstsq(
+        np.array(augmented, order="F", copy=True),
+        rhs,
+        lapack_driver="gelsy",
+        overwrite_a=True,
+        overwrite_b=False,
+    )
+    solve_seconds = time.perf_counter() - solve_started
+
+    def normal_relative(direction):
+        defect = (
+            jacobian.T @ (jacobian @ direction + r) + damping * direction
+        )
+        return float(np.linalg.norm(defect) / max(np.linalg.norm(gradient), 1e-30))
+
+    normal_svd = normal_relative(svd_direction)
+    normal_qr = normal_relative(qr_direction)
+    direction_difference = float(
+        np.linalg.norm(qr_direction - svd_direction)
+        / max(np.linalg.norm(svd_direction), 1e-30)
+    )
+    first = np.cos(np.arange(columns, dtype=float) + 0.25)
+    second = np.sin(np.arange(columns, dtype=float) + 0.75)
+    first /= np.linalg.norm(first)
+    second /= np.linalg.norm(second)
+
+    def dense_normal(direction):
+        return jacobian.T @ (jacobian @ direction) + damping * direction
+
+    first_image = dense_normal(first)
+    second_image = dense_normal(second)
+    normal_asymmetry = float(
+        abs(first @ second_image - second @ first_image)
+        / max(abs(first @ second_image), abs(second @ first_image), 1e-30)
+    )
+    minimum_probe_rayleigh = float(min(
+        first @ first_image,
+        second @ second_image,
+    ))
+
+    print("running the production-budget PCG comparator on stored J", flush=True)
+    jacobian_device = jnp.asarray(jacobian)
+    gradient_device = jnp.asarray(gradient)
+    pcg_result = pcg(
+        lambda direction: (
+            jacobian_device.T @ (jacobian_device @ direction)
+            + damping * direction
+        ),
+        -gradient_device,
+        rtol=1.0e-3,
+        atol=0.0,
+        max_steps=600,
+    )
+    jax.block_until_ready(pcg_result)
+    pcg_direction = np.asarray(pcg_result.x)
+    pcg_normal = normal_relative(pcg_direction)
+    admitted = bool(
+        rank == columns
+        and qr_rank == columns
+        and normal_svd < 1.0e-8
+        and normal_qr < 1.0e-8
+        and direction_difference < 1.0e-8
+        and normal_asymmetry < 1.0e-10
+        and minimum_probe_rayleigh > 0.0
+    )
+
+    if linear_data is not None:
+        direction_data = linear_data.with_name(
+            f"{linear_data.stem}-directions.npz"
+        )
+        np.savez(
+            direction_data,
+            gelsd_direction=svd_direction,
+            gelsy_direction=qr_direction,
+            pcg_direction=pcg_direction,
+        )
+        print("saved direct and PCG directions", flush=True)
+
+    initial_certificate = None
+    trials = []
+    accepted_fraction = None
+    final_certificate = None
+    if admitted:
+        print("linear gate passed; evaluating independent physics", flush=True)
+        initial_certificate = certificate(runtime.native)
+        rr, tt, zz = jnp.meshgrid(
+            jnp.asarray(runtime.radial_nodes),
+            jnp.asarray(runtime.theta),
+            jnp.asarray(runtime.zeta),
+            indexing="ij",
+        )
+
+        @jax.jit
+        def solve_grid_minimum_jacobian(vector):
+            state = _corrected_state(variable_scale * vector, runtime, chart)
+            samples = evaluate_strong_force(state, rr, tt, zz)
+            return jnp.min(runtime.native.jacobian_sign * samples.sqrt_g)
+
+        initial_cost = 0.5 * float(r @ r)
+        applied = jacobian @ svd_direction
+        gradient_dot = float(gradient @ svd_direction)
+        applied_squared = float(applied @ applied)
+        del augmented, jacobian_device, gradient_device
+        gc.collect()
+        for fraction in (1.0, 0.5, 0.25, 0.125, 0.0625):
+            candidate = jnp.asarray(fraction * svd_direction)
+            value = np.asarray(residual(candidate))
+            cost = 0.5 * float(value @ value)
+            predicted = (
+                -fraction * gradient_dot
+                - 0.5 * fraction**2 * applied_squared
+            )
+            ratio = (initial_cost - cost) / max(predicted, 1e-30)
+            state = _corrected_state(
+                variable_scale * candidate, runtime, chart
+            )
+            solve_grid_admissible = bool(
+                np.all(np.isfinite(value))
+                and float(solve_grid_minimum_jacobian(candidate)) > 0.0
+            )
+            trials.append({
+                "fraction": fraction,
+                "cost": cost,
+                "trust_ratio": ratio,
+                "solve_grid_admissible": solve_grid_admissible,
+            })
+            if solve_grid_admissible and ratio >= 1.0e-4 and cost < initial_cost:
+                accepted_fraction = fraction
+                final_certificate = certificate(state)
+                break
+
+    return {
+        "scope": (
+            "one dense direction at the initial production QA polish point; "
+            "fixed boundary and production chart/objective/scaling"
+        ),
+        "shape": [rows, columns],
+        "damping": damping,
+        "memory_bytes": memory,
+        "memory_budget_bytes": memory_budget,
+        "jax_xla_ad_memory_excluded_from_admission": True,
+        "setup_seconds": setup_seconds,
+        "jacobian_seconds": jacobian_seconds,
+        "factorizations_seconds": solve_seconds,
+        "gelsd_rank": int(rank),
+        "gelsy_rank": int(qr_rank),
+        "gelsd_original_normal_relative": normal_svd,
+        "gelsy_original_normal_relative": normal_qr,
+        "gelsd_gelsy_direction_relative": direction_difference,
+        "normal_operator_probe_asymmetry": normal_asymmetry,
+        "normal_operator_minimum_probe_rayleigh": minimum_probe_rayleigh,
+        "pcg_operator": "stored dense J.T@(J@x) + damping*x",
+        "pcg_linear_rtol": 1.0e-3,
+        "pcg_linear_max_steps": 600,
+        "pcg_iterations": int(pcg_result.iterations),
+        "pcg_reported_converged": bool(pcg_result.converged),
+        "pcg_reported_relative_residual": float(
+            pcg_result.relative_residual_norm
+        ),
+        "pcg_original_normal_relative": pcg_normal,
+        "pcg_gelsd_direction_relative": float(
+            np.linalg.norm(pcg_direction - svd_direction)
+            / max(np.linalg.norm(svd_direction), 1e-30)
+        ),
+        "linear_gate_passed": admitted,
+        "initial_certificate": initial_certificate,
+        "trials": trials,
+        "accepted_fraction": accepted_fraction,
+        "final_certificate": final_certificate,
+    }
+
+
 def experiment(native, layout, steps, order, angles, damping, planes=1):
     rho, weights = _span_quadrature(np.sqrt(native.radial_basis.breakpoints), order)
     # `planes` samples the field-period toroidal angle. One plane reproduces
@@ -177,7 +511,9 @@ def main():
     parser.add_argument("--order", type=int, default=6)
     parser.add_argument("--angles", type=int, default=24)
     parser.add_argument("--damping", type=float, default=1e-3)
-    parser.add_argument("--case", choices=("shaped", "qa"), default="shaped")
+    parser.add_argument(
+        "--case", choices=("shaped", "qa", "qa-production"), default="shaped"
+    )
     parser.add_argument("--planes", type=int, default=1)
     # Seed resolution is the representation under test: E2 requires at least
     # three radial refinements and two angular resolutions before a verdict.
@@ -186,6 +522,15 @@ def main():
     parser.add_argument("--seed-mpol", type=int, default=3)
     parser.add_argument("--seed-ntor", type=int, default=2)
     parser.add_argument("--budget-seconds", type=int, default=600)
+    parser.add_argument(
+        "--memory-budget-gib", type=float, default=2.0,
+        help="host dense-array admission cap; excludes JAX/XLA intermediates",
+    )
+    parser.add_argument("--jacobian-chunk-size", type=int, default=8)
+    parser.add_argument(
+        "--linear-data", type=Path,
+        help="optional private NPZ checkpoint of J, r, scales, and directions",
+    )
     args = parser.parse_args()
     if not jax.config.x64_enabled or not 1 <= args.steps <= 5 or not 4 <= args.order <= 12 or not 16 <= args.angles <= 64:
         parser.error("require float64, steps 1..5, order 4..12, angles 16..64")
@@ -205,6 +550,12 @@ def main():
         parser.error("seed resolution flags apply to --case qa; the shaped seed is fixed")
     if not np.isfinite(args.damping) or args.damping <= 0:
         parser.error("damping must be finite and positive")
+    if not np.isfinite(args.memory_budget_gib) or args.memory_budget_gib <= 0:
+        parser.error("memory-budget-gib must be finite and positive")
+    if not 1 <= args.jacobian_chunk_size <= 32:
+        parser.error("jacobian-chunk-size must be 1..32")
+    if args.case == "qa-production" and args.steps != 1:
+        parser.error("qa-production is one fixed-endpoint direction; use --steps 1")
     def expired(*_):
         raise TimeoutError(f"E2 {args.budget_seconds} s diagnostic budget")
     signal.signal(signal.SIGALRM, expired)
@@ -215,20 +566,43 @@ def main():
                       versions={name: version(name) for name in ("jax", "jaxlib", "scipy", "numpy", "solvax")},
                       device=str(jax.devices()[0]), precision="float64")
     started = time.perf_counter()
-    seed = {"shaped": ("shaped tokamak, frozen iota/pressure, GAMMA=0",
-                       shaped_state),
-            "qa": (f"nfp2 QA smooth beta at mpol={args.seed_mpol} ntor={args.seed_ntor} "
-                   f"ns={args.seed_ns} spans={args.seed_spans}, frozen iota/pressure, GAMMA=0",
-                   lambda: qa_state(mpol=args.seed_mpol, ntor=args.seed_ntor,
-                                    ns=args.seed_ns, spans=args.seed_spans))}[args.case]
-    output = {"_provenance": provenance, "case": seed[0],
-              "settings": {k: v for k, v in vars(args).items() if k != "output"}}
+    settings = {
+        key: value for key, value in vars(args).items()
+        if key not in ("output", "linear_data")
+    }
+    output = {"_provenance": provenance, "case": args.case,
+              "settings": settings}
     try:
-        output["result"] = experiment(*seed[1](), args.steps, args.order, args.angles,
-                                      args.damping, args.planes)
+        if args.case == "qa-production":
+            output["result"] = production_qa_discriminator(
+                args.damping,
+                args.memory_budget_gib,
+                args.jacobian_chunk_size,
+                args.linear_data,
+            )
+        else:
+            seed = {
+                "shaped": ("shaped tokamak, frozen iota/pressure, GAMMA=0",
+                           shaped_state),
+                "qa": (
+                    f"nfp2 QA smooth beta at mpol={args.seed_mpol} "
+                    f"ntor={args.seed_ntor} ns={args.seed_ns} "
+                    f"spans={args.seed_spans}, frozen iota/pressure, GAMMA=0",
+                    lambda: qa_state(
+                        mpol=args.seed_mpol, ntor=args.seed_ntor,
+                        ns=args.seed_ns, spans=args.seed_spans,
+                    ),
+                ),
+            }[args.case]
+            output["case"] = seed[0]
+            output["result"] = experiment(
+                *seed[1](), args.steps, args.order, args.angles,
+                args.damping, args.planes,
+            )
         output["status"] = ("completed bounded diagnostic at this resolution; "
                             "nonlinear convergence and production promotion not established")
-        if not output["result"]["verification_passed"]:
+        if (args.case != "qa-production"
+                and not output["result"]["verification_passed"]):
             raise RuntimeError("E2 reference consistency or geometry check failed")
     except Exception as error:
         output["status"] = f"{type(error).__name__}: {error}"
```

</details>

rogeriojorge:
Fold into Part I before merging (independent review of the derivative stack, full record on #417: https://github.com/uwplasma/vmex/pull/417#issuecomment-5766740229):

- #416 merged (`6f1df72`): regression fails on main and passes with the change; free-boundary file 28 passed / 1 skipped with RUN_FULL=1.
- #417 premise did NOT reproduce on main. Fixed boundary (li383_low_res, 8 dofs, ftol 1e-13): Jacobians identical to #424, gradient agrees with cold re-solve finite differences to 5.3e-4. Free boundary (asymmetric deck, ns=8, ftol 1e-12): adjoint agrees with independent re-solves to 1.3-1.8e-4 (pressure) and 1.5-4.3e-4 (coil current). The stack does not touch the free-boundary module. Recommendation: do not merge the +1200/-157 stack as is.
- Real defect 1: mean-iota implicit derivative differs from stable re-solve finite differences by 0.9-1.2% on RBC(1,1)/ZBS(1,1), up to ~20% on the small ZBS(-1,1) entry, on main and on the stack. The three Jacobian methods agree with each other to 5 digits, so the gap is in the linearization, consistent with the frozen m=1 family noted here. Taylor checks on `wb` cannot see it; add an iota-based check.
- Real defect 2: `equilibrium_from_x` returns the unrefined state; on li383 mean iota is 0.55359 in the objective vs 0.55449 in the written output. Fix as one small PR onto main (return the refined state + a test comparing objective iota to written iota), then close #417/#418/#421/#423/#424 pointing to it.
- Stack CI diagnosis: c3d (#417/#421/#423) = `equilibrium_from_x` before any certificate (fixed by #424); Quality (#421) = mypy call-arg on `measure_primal_state`; implicit-response-b (#423/#424) = exact 0.0 assertion gets 2.4e-17; a-core (#424) = HTTP 504 flake.
- Per-gradient cost of the admission checks: not measured.

PR #416, opened 2026-09-21
Keep free-boundary cold recovery deterministic

rogeriojorge:
## What this changes

A free-boundary trial whose fixed-reference restart missed `ftol` used a global per-configuration cold-rebuild budget. After unrelated rejected trials exhausted that budget, evaluating the same recoverable point returned the stalled restart instead of its converged cold root. This made the reported state and status depend on trial history.

Every stalled restart now gets exactly one deterministic cold retry. The fixed reference remains unchanged, and a failed cold retry remains a status-2 trial. This preserves the root equation and makes repeated accepted points stable across unrelated failures. The tradeoff is explicit: after the former eight-failure limit, each rejected point may again pay for one cold retry; suppressing that retry based on shared history was incompatible with the parameter-to-root contract.

## Evidence

- A focused regression evaluates one recoverable point, exhausts the former budget with eight unrelated status-2 trials, and requires the repeated point to return status 0 with bit-identical state. It failed on the parent revision and passes here.
- The real asymmetric free-boundary case evaluates an accepted current, an unrelated doubled-current status-2 trial, then the accepted current again. The accepted roots agree bit for bit.
- `python tools/preflight.py`: PASS; 20 affected tests passed, 86 guard tests passed, and static checks passed.
- Changed executable coverage: 100% (2/2 lines).
- Full physical test: 1 passed in 90.35 s on CPU.

This does not change the equilibrium equations, seeds, objective weights, or acceptance tolerances. It does not overlap the surface-field fix in #409 or the example/reporting work in #377.


rogeriojorge:
### Independent pressure re-solve check

I tested the recovery change together with the exact-state anchor work in #417. The exact source revisions were `85517771` and `559b1f99`, respectively. The run used Python 3.12.13, JAX 0.11.1 with 64-bit values enabled, NumPy 2.5.3, SciPy 1.18.1, and an x86-64 Intel Xeon W-2295 CPU.

The cache-cleared base and repeat were bit-identical (`J = 0.27524957412450995`) and each finished in 95 iterations. Thus the recovery change removes trial-history variation for this protocol. The returned base state nevertheless measured:

- native summed FSQ `1.4791836e-9` (`FSQ / ftol = 1.479`; native status 0 because each component met `ftol`);
- projected/preconditioned coupled residual norm `9.1454448e-6`;
- raw coupled residual norm `5.6087822e-4`.

For `J = mean(R_edge^2 + Z_edge^2)` in the `am[0]` scaling direction, the one-root adjoint was `1.7726117e-5`. Independent cache-cleared central re-solves gave:

| relative step | re-solve derivative | relative difference from adjoint |
|---:|---:|---:|
| `1e-2` | `1.5659175e-5` | `11.66%` |
| `1e-3` | `1.5201522e-5` | `14.242%` |
| `1e-4` | `1.5201631e-5` | `14.242%` |

The last two re-solve derivatives agree to about `7.2e-6` relative, while all legs retain projected residuals near `9.14e-6` and raw residuals near `5.61e-4`. The finite difference is therefore stable and deterministic but does not approach the one-root derivative. This supports a remaining nonlinear-root accuracy limitation; it does not indicate an adjoint operator inconsistency by itself. A bounded follow-up at tighter native tolerances will distinguish those cases.

Executable reproduction from a clean checkout:

```bash
git switch --detach 559b1f997ef4cc92946b5cf516ed49b76846d20d
git cherry-pick 85517771d80d514a7e62dab2e57e7ccba8fa25f6
JAX_ENABLE_X64=True python - <<'PY'
import dataclasses
from pathlib import Path
import jax
import jax.numpy as jnp
import numpy as np
from tests.test_lasym_free_case import lasym_free_field, lasym_free_input
from vmex.core import freeboundary_implicit as fbi
from vmex.core import implicit as im

data = Path("examples/data")
inp = dataclasses.replace(
    lasym_free_input(data), ns_array=np.array([8]),
    ftol_array=np.array([1e-9]), niter_array=np.array([4000]))
field = lasym_free_field()
params = im.params_from_input(inp)
cfg = fbi.make_free_boundary_config(
    inp, field, ns=8, ftol=1e-9, max_iterations=4000,
    adjoint_tol=1e-7, adjoint_maxiter=150,
    field_from_parameters=lambda current: dataclasses.replace(
        field, extcur=current), device="cpu")

def objective(state):
    return jnp.mean(state.R_cos[-1]**2 + state.Z_sin[-1]**2)

def cold(parameters):
    fbi._FREE_HOT_CACHE.pop(cfg, None)
    (state, status, fsq, ratio), saved = fbi._solve_status_fwd(
        parameters, field.extcur, cfg)
    prm, current, solved, mask, rcon0, zcon0, _ = saved
    project = im._dof_projector(cfg.implicit, mask)
    frozen, z = jax.lax.stop_gradient(solved), project(solved)
    norms = []
    for form in ("preconditioned", "raw"):
        residual = fbi._projected_residual(cfg, mask, formulation=form)
        norms.append(float(im._tree_norm(
            residual(z, prm, current, frozen, rcon0, zcon0))))
    result = fbi._FREE_LAST_RESULT[cfg]
    return float(objective(state)), int(status), float(fsq), float(ratio), \
        int(result.iterations), norms, saved

base = cold(params)
prm, current, solved, mask, rcon0, zcon0, _ = base[-1]
direction = dataclasses.replace(
    jax.tree.map(jnp.zeros_like, prm),
    am=jnp.zeros_like(prm.am).at[0].set(prm.am[0]))
state_bar = jax.grad(objective)(solved)
params_bar, _ = fbi._solve_bwd_impl(
    cfg, (prm, current, solved, mask, rcon0, zcon0), state_bar)
adjoint = float(sum(jnp.vdot(a, b) for a, b in zip(
    jax.tree.leaves(params_bar), jax.tree.leaves(direction))))
print("base", base[:-1])
print("repeat", cold(params)[:-1])
print("adjoint", adjoint)
for step in (1e-2, 1e-3, 1e-4):
    values = []
    for sign in (-1.0, 1.0):
        shifted = jax.tree.map(
            lambda value, delta: value + sign * step * delta,
            params, direction)
        values.append(cold(shifted))
    finite = (values[1][0] - values[0][0]) / (2 * step)
    print(step, finite, abs(finite - adjoint) / abs(adjoint),
          values[0][1:-1], values[1][1:-1])
PY
```

rogeriojorge:
### Tighter-root follow-up

Using the same pinned revisions and environment as the reproduction above, I repeated the study with `ftol=1e-10`, `max_iterations=8000`, and central-difference steps `1e-3` and `1e-4`.

The independently cache-cleared base and repeat now agree in every state coefficient (`max(abs(state_repeat - state_base)) = 0`) as well as in the objective. The base took 100 iterations and measured native summed FSQ `1.2631561e-10`, projected coupled residual `1.7399623e-6`, and raw coupled residual `1.5586784e-4`.

The one-root adjoint was `1.6992190e-5`. Independent re-solves gave `1.5541122e-5` at `h=1e-3` and `1.5541201e-5` at `h=1e-4`; those differences agree to about `5.1e-6` relative and remain `8.54%` below the adjoint.

Relative to the `ftol=1e-9` run, the projected residual fell by 5.26x, the raw residual by 3.60x, and the derivative discrepancy narrowed from 14.24% to 8.54%. This trend supports incomplete nonlinear-root accuracy as a material cause, but does not establish that the operator is consistent in the asymptotic limit. The next bounded run tightens `ftol` again and stops if the achieved root plateaus or reaches its iteration cap.

To reproduce this point, use the preceding executable block with these substitutions and add the coefficient comparison after its base repeat:

```python
ftol_array=np.array([1e-10])
niter_array=np.array([8000])
# make_free_boundary_config(..., ftol=1e-10, max_iterations=8000, ...)
# for step in (1e-3, 1e-4):
repeat = cold(params)
base_state, repeat_state = base[-1][2], repeat[-1][2]
coefficient_delta = max(
    float(np.max(np.abs(np.asarray(a) - np.asarray(b))))
    for a, b in zip(jax.tree.leaves(base_state),
                    jax.tree.leaves(repeat_state)))
```

rogeriojorge:
### Asymptotic pressure derivative reached

The final bounded point used #417 revision `55327c99` plus recovery revision `85517771`, with `ftol=1e-12`, `max_iterations=8000`, and the same independently cache-cleared protocol. The base and repeat agree in every state coefficient (`max(abs(state_repeat - state_base)) = 0`) and in the objective. The base finished in 114 iterations with native summed FSQ `1.5433196e-12`, projected coupled residual `1.9667032e-7`, and raw coupled residual `1.7173762e-5`.

The one-root adjoint was `1.66451415e-5`. Independent re-solves gave:

| relative step | re-solve derivative | relative difference from adjoint |
|---:|---:|---:|
| `1e-3` | `1.66481015e-5` | `1.778e-4` |
| `1e-4` | `1.66479203e-5` | `1.669e-4` |

The two finite differences agree to about `1.09e-5` relative. The complete bounded progression is:

| `ftol` | iterations | projected norm | raw norm | adjoint | FD (`1e-3`) | FD (`1e-4`) | small-step gap |
|---:|---:|---:|---:|---:|---:|---:|---:|
| `1e-9` | 95 | `9.1454e-6` | `5.6088e-4` | `1.77261e-5` | `1.52015e-5` | `1.52016e-5` | `14.24%` |
| `1e-10` | 100 | `1.7400e-6` | `1.5587e-4` | `1.69922e-5` | `1.55411e-5` | `1.55412e-5` | `8.54%` |
| `1e-12` | 114 | `1.9667e-7` | `1.7174e-5` | `1.66451e-5` | `1.66481e-5` | `1.66479e-5` | `0.0167%` |

The adjoint discrepancy falls as the measured projected and raw residuals fall. This establishes an independent re-solve asymptotic window for this pressure observable and deck. It attributes the earlier discrepancy to incomplete nonlinear-root accuracy here; it is not a claim about every free-boundary observable or equilibrium family.

No `ftol=1e-14` run was started because the requested diagnostic question is resolved at `1e-12`: both step sizes agree with each other and with the one-root adjoint, and the solve has neither plateaued nor reached its iteration cap.

The executable reproduction in the first comment applies with these substitutions:

```python
ftol_array=np.array([1e-12])
niter_array=np.array([8000])
# make_free_boundary_config(..., ftol=1e-12, max_iterations=8000, ...)
# for step in (1e-3, 1e-4):
```

PR #417, opened 2026-09-21
Keep implicit derivatives anchored to measured states

rogeriojorge:
## Problem and change

A native solve could report convergence while refinement returned a different or insufficiently resolved state, and optimizer Jacobian reuse was keyed only by design parameters. That allowed an objective residual at one state to be paired with a response from another.

This change measures projected/preconditioned and raw nonlinear residual norms, normalized raw FSQ, coefficient finiteness, and geometry on the exact returned refined coefficients. It records parameter and coefficient identities, requires that provenance for optimizer admission and Jacobian/predictor reuse, rechecks it after forward or reverse response work, and materializes the refined state. `primal_tol=None` avoids a universal absolute residual cutoff; callers may provide an observable-specific finite bound. Diagnostics separately report derivative admission, requested-refinement success, and strict root certification. Direct implicit workflows can obtain the same exact-state evidence through `measure_primal_state` without inferring native convergence. Scalar reverse objectives retain a single `J^T r` pullback.

## Validation

- Focused implicit-gradient, optimizer, penalty, diagnostics, and independent free-boundary pressure suite: 21 passed on the rebased source.
- Scalar reverse fast-path and unrelated `ValueError`/`TypeError`/`JaxRuntimeError` propagation regression: passed.
- Changed executable line coverage against `main`: 98% (319/325).
- A difficult public asymmetric case reduced the projected residual from `2.67e-7` to `4.77e-15` while moving the state by `1.80e-3`; explicit tight and loose admission bounds reject and accept that measured state as expected.
- The same physical case exercises the public direct-state measurement accessor and verifies that default admission is not labeled strict or native convergence.
- A small physical Solovev case checks block, GMRES, and reverse agreement; repeating the original point after a rejected trial; and rejection of synthetic state drift during a response.
- Independent re-solves at three perturbation sizes give directional Taylor orders above 1.7 for a physical observable.
- On the free-boundary pressure case, independent re-solves at `ftol=1e-12` agree with the one-root adjoint to about 0.017% at two step sizes; looser roots give stable but biased differences. The bounded three-tolerance evidence and reproduction are recorded on #416.
- Force-resolution benchmark outputs now carry the public exact-state measurement instead of calling a seed natively converged without native status evidence.
- On the small warmed Solovev certificate kernel, adding the raw residual measurement increased the median time from 0.191 ms to 0.200 ms (about 5%). This is a bounded CPU microbenchmark, not a general performance claim.
- Ruff, byte compilation, and `git diff --check` pass.

## Limits

The injected failure tests establish routing and provenance behavior; they do not establish physical correctness. The physical cases support the discrete linearization, state anchoring, and one observable's Taylor behavior. They do not establish equilibrium uniqueness, continuum-force convergence, every objective or geometry, full-optimization performance, GPU performance, or process-global cache isolation across concurrent solves. With the default `primal_tol=None`, admission still requires finite coefficients and residuals, valid geometry, and the configured raw-FSQ ratio, and diagnostics expose an unmet refinement target without calling it a strict root certificate.


rogeriojorge:
## Independent review handoff (stack #417 -> #418 -> #421 -> #423 -> #424)

**Premise: not reproduced on main. Gradients were already correct in normal use. This stack is defensive hardening, not a fix for a wrong-gradient bug.**

Fixed boundary (`li383_low_res`, rows are aspect ratio and mean iota, 8 boundary dofs, ftol 1e-13). Each case compares the optimizer Jacobian from `make_problem` with central differences of independent cold re-solves.
- The `main` Jacobian and the #424 Jacobian are bit-identical (max |ΔJ| = 0.0). Their `value_and_grad` gradients differ by 1.6e-13.
- History independence on `main`: revisiting x0 after an accepted nearby step, or after a rejected far trial (penalty residual 2.6e7), returns the same residual to 1e-16 and the same Jacobian to 1.3e-13. No residual/Jacobian pairing across states was observed.
- `J^T r` agrees with FD to 5.3e-4. The aspect-ratio row agrees to about 1e-8.

Free boundary (asymmetric lasym deck, ns=8, ftol 1e-12). On `main`, the adjoint agrees with independent cold re-solves at h = 1e-3, 1e-4 and 1e-5 as follows:
- Pressure `am[0]`: 1.3e-4 to 1.8e-4.
- Coil `extcur[0]`: 1.5e-4 to 4.3e-4.

The new physical tests in this stack both pass unchanged against `main` code: `test_free_boundary_pressure_gradient_matches_tight_independent_resolves` and `test_solovev_directional_taylor_with_independent_resolves`. The stack never touches the free-boundary derivative path, because `freeboundary_implicit` calls none of the new admission code. Its free-boundary cost and benefit are therefore both zero.

**Real problem found, and this stack does not change it:** the mean-iota derivative is off relative to independent re-solves.
- The block, tight-block and reverse Jacobian methods agree with each other to 5 digits.
- FD of refined re-solves is stable across h = 2e-3, 1e-3 and 5e-4, and gives identical values at ftol 1e-13 and 1e-15.
- The implicit value differs from that FD by 0.9–1.2% on RBC(1,1) and ZBS(1,1), and by up to about 20% on the small ZBS(-1,1) entry.
- State-level tangents differ by O(1). This matches the frozen-residual root family noted in #413 and #377.
- The `wb` Taylor test cannot detect this, because `wb` is variational. Use a non-variational observable such as iota for derivative evidence.

**Real, small defect this stack does fix (#424):** on `main`, `equilibrium_from_x` returns the native, unrefined state, not the refined state the objective used. On `li383_low_res` at default settings the two states differ by up to 1.4e-2 in a coefficient. The objective's mean iota is 0.55359 while the materialized WOUT gives 0.55449, a 0.16% mismatch. A minimal PR could fix only this: materialize `_LAST_REFINED`, and refresh through `_host_solve_and_mask_status` when it is missing.

**CI failures, diagnosed but not fixed (no pushes were made):**
- #417, #421 and #423, lane c3d: `test_equilibrium_exterior_field_sizes_its_source_grid` raises "did not produce a certified VMEC equilibrium". It calls `equilibrium_from_x` straight after construction, when no certificate exists yet. #424 fixes this.
- #421, Quality: mypy reports `Too many arguments for "measure_primal_state"` at `implicit.py:2035`. Later commits in the stack change that call.
- #423 and #424, implicit-response-b: `test_primal_boundary_gate_rejects_a_perturbed_supplied_edge` asserts an exact `0.0` but gets 2.43e-17. The fix is to compare against the roundoff allowance that `_primal_boundary_error` already returns.
- #424, a-core: an HTTP 504 during setup. This is infrastructure; rerun it.
- c2 is cancelled at its cap on every PR until #407 lands.

**Not measured:** the per-gradient cost of the new admission checks. Each status solve adds a fresh force evaluation plus a SHA-256 of the state. Every direct bwd, tangent or pullback adds another force evaluation, and each new config adds one jit compile. The A/B/A/B benchmark was stopped before it produced numbers.

**Recommended next steps:**
1. Do not merge the stack as it stands: +1200/-157 across 16 files, with no reproduced defect. It also renames the public `derivative_certified` diagnostic.
2. Open one small PR onto `main` that makes `equilibrium_from_x` materialize the refined state, with a regression test comparing objective iota to WOUT iota.
3. If exact-state evidence is still wanted, return the final projected `|F|` from `_refined_state`. It is already computed on the returned coefficients, so this costs nothing and avoids measuring the force again.
4. Track the iota frozen-family derivative gap separately (#377 line of work).
5. Close #417, #418, #421, #423 and #424 with a pointer to that PR.

rogeriojorge:
## Free-boundary adjoint audit and the mean-iota gap (follow-up to the review above)

**Scope and conditions**
- Code:
  - `main` at 780eb86eb (0.11.0).
  - The single-stage example is `single_stage_free_boundary_optimization_finite_beta.py` from #411 at 08d97509f: 0.5% beta, ns 31, mpol = ntor = 5, 81 coil dofs.
  - Its optimized coils come from a #411 run that met every target.
- Hardware: every number below was measured on the laptop under load from other jobs (load average 30-230). The office machine became unreachable mid-run, so all timings are labelled "under load".
- Timing protocol: a pinned, clean worktree; value and value+gradient alternated V/G/V/G at distinct points.

### Verdict

- **The free-boundary adjoint is exact.** At the Newton-anchored root of the projected coupled residual it linearizes, it matches central FD of re-rooted states with the same frozen complement to 1e-9 to 6e-7 at h = 1e-3 to 3e-4. This holds at the seed and at the optimum, for the objective and for every term.
- **The values the optimizer sees are not at that root.**
  - A converged solve at ftol 1e-12 (status 0) sits 1.15e-2 (seed) and 1.26e-2 (optimum) from the root in coefficient norm. Four Newton steps take |F| from 3e-7 to 5e-14.
  - Warm restarts from the seed reference stop after moving part of the way.
  - Which solve path runs changes the value by up to 12%.
  - So gradient and values disagree at the scale of a line-search step. The gradient itself is not wrong.
- **The mean-iota gap is the released m=1 constrained family**, not a missing linearization term (details below and in #428).

### 1. Accuracy (relative error vs central FD)

Free boundary, seed coils, ftol 1e-12, direction = normalized objective gradient:

| term | vs frozen-path FD of anchored roots (h = 3e-4) | vs anchored cold re-solves | vs raw cold re-solves (plateau, h = 1e-2 to 1e-3) | vs production warm re-solves (h = 1e-3) |
|---|---|---|---|---|
| objective | 6.0e-8 | 4.0e-4 | 6.0e-3 | 0.94 |
| QS 0.5 r^2 | 2.6e-7 | 1.0e-2 | 1.9-2.6e-2 | 1.9 |
| beta | 8.2e-9 | 5.8e-3 | 1.9-3.9e-3 | 0.25 |
| mean iota | 1.8e-8 | 6.8e-4 | 1e-5 to 1.1e-3 | 0.94 |
| min iota | 2.8e-8 | 3.8e-4 | 6.0e-3 | 0.94 |
| aspect | 1.0e-7 | 1.3e-2 | 1.6e-2 | 0.40 |

- At the optimized coils (h = 1e-3), the frozen-path column is 1.3e-9 to 6.4e-7 for every term, including the objective.
  - At h = 1e-2 the objective reads 5e-2, and 1.2e-2 at h = 3e-3. That error falls in proportion to h: the FD step crosses a hinge (active coil-distance and curvature penalties).
  - The anchored-cold column at the optimum is 2.5e-6 to 6.8e-3.
- A random direction at the seed gives the same pattern: frozen-path 4e-9 to 1.3e-4 at h = 1e-2; anchored-cold 3e-3 to 3.7e-2.
- Raw cold FD in two other directions (a single dof, and a second random vector) reaches 1-10% on QS and 22-34% on small aspect components. The warm FD varies with h by factors of 2-20 in every direction.

**Why warm FD fails.** Along the gradient direction at the seed:
- At h = 3e-4, the warm restart moves the state 5.5e-5, against 2.6e-4 for the cold solve.
- At h = 1e-3, dJ/h reads 15.6 from warm solves, against 30.0 from cold solves and 29.7 from the adjoint.

The restart meets ftol along a soft mode before it reaches the root. The fixed-boundary lane anchors against exactly this with `_refined_state`; the free-boundary lane has no anchor.

Fixed boundary, QA seed deck of the same example (fixed boundary, 0.5% beta, ns 31, ftol 1e-13), cold (refined) re-solve FD at h = 3e-4, stable from 1e-3 to 1e-4:

| dof | QS | beta | mean iota | min iota | aspect |
|---|---|---|---|---|---|
| RBC(1,1) | 1.3e-3 | 5.1e-5 | 2.0e-4 | 9.3e-4 | 7e-7 |
| ZBS(1,1) | 2.4e-3 | 5.1e-5 | 2.0e-4 | 9.2e-4 | 7e-7 |
| RBC(0,2) | 1.5e-5 | 1.0e-5 | 3.8e-6 | 8.1e-2 (profile minimum) | 3e-13 |
| ZBS(-1,1) | 1.4e-3 | 7.7e-3 | 3.2e-2 | 2.1e-1 | (derivative 0) |
| PRES_SCALE | 1.3e-5 | 4.6e-9 | 8.2e-8 | 2.1e-6 | 0 |

- The pressure derivatives are exact.
- The boundary rows carry the m=1 family term described in section 3.

### 2. Speed and memory (under load; pinned worktree, clean)

| case | dofs | value | value + gradient | ratio | forward FD (n values) |
|---|---|---|---|---|---|
| free boundary, coil order 2, near seed | 45 | 0.58 s | 6.2 s | 10.8 | 26 s (4x) |
| free boundary, coil order 4, near seed | 81 | 0.92 s | 7.8 s | 8.4 | 75 s (10x) |
| free boundary, coil order 6, near seed | 117 | 1.08 s | 10.0 s | 9.2 | 126 s (13x) |
| free boundary, coil order 4, at optimum | 81 | 29.8 s | 35.3 s | 1.19 | 2415 s (68x) |
| fixed-boundary single stage, coil order 4 | 105 | 5.7 s | 11.6 s | 2.0 | 594 s (51x) |

- The gradient adds 5.5-9 s, almost independent of the number of coil dofs. It beats FD by 4-13x near the seed and by 68x mid-optimization.
- The high ratio near the seed means the value is cheap there (a short restart from the reference), not that the adjoint is expensive.
- Memory:
  - Live JAX arrays stay constant (210 arrays, 1.2 MiB) over 20 consecutive gradients, so there is no per-gradient leak.
  - Peak RSS is 2.4-3.2 GiB for the trial loop, and 4.5 GiB for fixed boundary.
  - macOS RSS under memory pressure fluctuates ±0.5 GiB, so it is not a leak signal.

**Wasted restart work.** Most of the cost at the optimum is a restart that cannot converge. Tracing the stages of a production trial at t·u_opt:

| t | seed-reference restart | cold fallback | trial |
|---|---|---|---|
| 0.10 | 329 iterations, converged | none | 1.7 s |
| 0.25 | 1,305 iterations, converged | none | 6.2 s |
| 0.5 / 0.75 / 1.0 | 4,000 iterations (cap), not converged, 19-21 s | 823 / 1,287 / 1,425 iterations, 4-7 s | 25 s |

- In the second half of an optimization, every trial spends about 80% of its time in a restart that hits the cap.
- A restart budget or stall test before the cold fallback would cut the trial cost about 4x.

### 3. Mean-iota root cause (fixed boundary suffices)

The implicit derivative holds the released m=1 `Z_sin` pair combination at its converged value. VMEC zeroes its force once `fsqz < 1e-6`. Each cold re-solve freezes that combination wherever its own path left it, at about 0.5 per unit boundary change. On `li383_low_res` (ftol 1e-13, h = 5e-4):

| direction | implicit | re-solve FD | gap | frozen-path FD | iota response to the family drift only | implicit + response vs re-solve |
|---|---|---|---|---|---|---|
| RBC(1,1) | -2.551476 | -2.526064 | 1.0% | 6.6e-7 | +0.025418 | 2.3e-6 |
| ZBS(-1,1) | +0.787230 | +0.766046 | 2.8% | 3.6e-6 | -0.021172 | 1.6e-5 |

- **(a) Missing term:** no. The linearization is exact for the gauge-fixed equations.
- **(b) Iota-evaluation inconsistency:** no. Each re-solve state is exactly a root of the frozen residual with its own frozen entries (residual 1e-14).
- **(c) FD artifact:** no. The re-solve FD is stable across h and ftol.
- The coordinate is an angle gauge in the continuum.
  - Over mpol 4-8 and ns 16-64 the gap spans 0.06-1.0% (RBC(1,1)) and 0.1-2.8% (ZBS(-1,1)), and it does not fall monotonically. The finest case, mpol 8 / ns 32, gives 0.27% and 0.87%.
  - The free-boundary "anchored cold" column in section 1 is the same effect.
- Literature:
  - The continuous-adjoint VMEC papers (Antonsen, Paul & Landreman 2019; Paul et al. 2020, 2021) work in the continuum, where the angle perturbation drops out. They report agreement "within several percent", attributed to discretization.
  - DESC fixes only the lambda gauge and does not constrain the poloidal angle. Its perturbation derivatives come from the Jacobian's least-squares solve and also hold that freedom fixed.

### 4. Status determinism at finite beta

At ftol 1e-10, t·u_opt was evaluated for t = 0.25, 0.5, 0.75 and 1.0, each from three starts: the seed reference, a reference at 0.5 u_opt, and a cold solve.
- Every evaluation returned status 0, so no status flip was seen at finite beta.
- The value did depend on the start:
  - t = 0.25: J = 1.1598 (seed reference), 1.3257 (mid), 1.2959 (cold). States differ by 2e-2.
  - t = 0.75: 0.1749, 0.1660 and 0.1752.
  - t = 1: 0.028896 (seed reference, which is the cold-fallback branch) against 0.028728 (cold then restart).
- The production map switches from the restart branch to the fallback branch between t = 0.25 and 0.5, and J jumps there.
- So a gradient is not taken at a status-2-on-resolve point at finite beta, but it is taken on a value map that is not a function of u alone across references. The #411 re-solve with a new configuration does not reproduce the trial values.

### 5. This stack (#417/#418/#421/#423/#424)

- None of it touches `freeboundary_implicit.py` or the m=1 handling, so it changes none of the numbers above.
- The one real defect it fixes, `equilibrium_from_x` returning the unrefined state, is now #427 on `main`.
  - It is a 12-line fix with a regression test: objective iota 0.553590 vs WOUT iota 0.554491 on `main`, identical with the fix.
  - Its CI is green.
- #428 adds the mean-iota closure test and documents the cause. Its CI is green.
- Neither is merged.

### Recommended next steps (not done here; they touch free-boundary solver code another agent is working in)

1. Anchor the free-boundary state with a Newton refinement on `_projected_residual`, as `_refined_state` does for fixed boundary. Measured: 4 steps, |F| from 3e-7 to 5e-14, so value and gradient would refer to the same point.
2. Give the reference restart a stall or iteration budget before the cold fallback. At the optimum that is about 20 s of each 25 s trial.
3. If sub-percent agreement with re-solves is wanted for iota, pin the m=1 family to a canonical, parameter-differentiable value (for example the interior-guess value) instead of the path value.

rogeriojorge:

PR #418, opened 2026-09-21
Validate direct derivative states without an absolute tolerance

rogeriojorge:
Stacked on #417 at `1459f9df54d2e1c7990975b1c1aba1621681c2fa`. This focused follow-up does not change the equilibrium equations or introduce a universal primal tolerance.

Direct implicit derivative calls returned an unconditional true admission predicate when `primal_tol=None`, bypassing the fresh residual, geometry, and FSQ checks added by #417. An invalid refined geometry or nonfinite residual could therefore reach a direct response path while the optimizer/status path rejected it.

This change always runs the shared eligibility measurement. `None` continues to disable only the absolute projected-residual cutoff. Eager invalid states raise `AdjointSolveError`; staged responses receive false eligibility and use the existing NaN guard. The error tolerance is `None` when admission fails and no absolute norm bound was configured.

The restack preserves #417's measured-state accessor, serialization changes, independent free-boundary response check, and benchmark reporting. The resulting head is `9e0baa287dac70be56f4c436225fca11f11a9ccd`.

Validation on the exact stacked tree:

- `tests/test_optimize_penalty.py`: 18 passed in 69.97 s on the exact restacked tree with Python 3.11.14, JAX/JAXlib 0.10.1, CPU-only execution, one CPU thread, and JAX 64-bit mode enabled.
- Five parameterized cases cover invalid geometry, nonfinite projected and raw residuals, negative FSQ, and excess FSQ in eager and staged execution.
- A large finite projected residual remains eligible when `primal_tol=None` and all unconditional checks pass; a configured finite bound still rejects an excessive norm.
- The staged regression uses a traced state and verifies false eligibility plus the existing NaN propagation.
- `git diff --check` passes.

Reproduce the focused module with:

```sh
JAX_ENABLE_X64=1 PYTHONPATH=. python -m pytest -q tests/test_optimize_penalty.py
```

This adds a fresh force measurement to direct backward and multi-RHS admission. Status-based responses can repeat a measurement already performed in forward certification; safely carrying that evidence through staged computation is a separate performance opportunity. No new performance, GPU, or continuum force-balance claim is made. Keep this draft until the parent and this correction are reviewed together.


rogeriojorge:
I validated this patch without modifying the contributor branch by applying commit `7b7156920144a69ac569b86355cccd69be5d8f03` onto the current PR #417 head `1459f9df54d2e1c7990975b1c1aba1621681c2fa`.

The three-file patch applied cleanly and preserved PR #417's direct-gate coverage. Focused eager/JIT validation passed:

```
JAX_DISABLE_JIT=0 python -m pytest -q tests/test_optimize_penalty.py -k 'direct_derivative_always_checks_primal_validity or status_callback_uses_configured_actual_state_tolerance'
8 passed, 10 deselected in 2.08s
```

Please restack this branch onto `1459f9df54d2e1c7990975b1c1aba1621681c2fa` so the published history matches the jointly validated source.

rogeriojorge:

PR #421, opened 2026-09-21
Align host callback refinement with its runtime device

rogeriojorge:
Host callbacks receive CPU payloads even when the numerical root or cached runtime belongs to another device. Refinement could mix those arrays and fail before producing a usable result. This change places callback-local parameters, state and mask together on the explicit configured device or the cached runtime device, without changing callback sharding or public configuration/cache keys.

Stacked on #418 (`9e0baa287dac70be56f4c436225fca11f11a9ccd`), which follows #417. The integrated certificate path measures the exact returned coefficients on that same runtime device; it never substitutes a cached refined state. Certificate provenance remains keyed to the returned state. Review and merge in dependency order.

#422 is merged into this PR as `e94e46c6c11f98dd278f7647fbf5d84761a80fce`. It adds complementary placement tests only; no second production implementation remains. Independent review found no blocker. At the exact test head `e25aa1fd`, all 11 cases passed with two forced CPU devices, including an explicit `_module_jit_enabled` run (1.31 s). The default test command disables JIT and was measured separately (9.38 s). Main integration still requires this consolidated branch's checks/review and actual GPU-root qualification.

An initial implementation used a nonexistent RunSetup.grids attribute. A real numerical solve exposed it; the correction uses the actual setup.s_full field, with a real runtime fixture rather than an invented mock attribute.

Validation of the final source and test bytes at `95aed755`:

- Device and optimizer-penalty modules: 52 passed, 5 skipped.
- Focused tests with two forced CPU devices: 5 passed. The new certification regression fails on the parent for CPU1 placement and passes after the change.
- Real small CPU callback plus fresh certificate: status 0, finite result, JIT enabled, 15.86 seconds. The public Solovev case uses MPOL3, NTOR0, NS5 and explicit bounded resource settings.
- The certificate regression places a real runtime on each CPU device and supplies a deliberately different cached refined state, checking both device placement and exact returned coefficients.
- Static checks and independent source review passed. No numerical files changed after the recorded tests; rebasing preserved verified file hashes.

Qualification limits:

- The earlier small GPU-facing check at `6abcdbac` performed its callback/root/refinement work on CPU; it does not establish accelerator-resident root correctness or CPU/GPU parity.
- One bounded cached-GPU-root attempt at that same revision stopped at 120 seconds during root priming, before callback entry. No GPU root or certificate was obtained. This does not identify a deadlock or isolate a specific performance cause.
- The earlier full QA explicit-GPU capture timed out at 660 seconds without output arrays. No full QA GPU result is claimed.
- Explicit configured-device precedence over a disagreeing cached template is not directly covered by the new test.
- Complete optimization, exact-case QA and accelerator-resident callback qualification remain open. Keep this PR draft until the intended integration gates are satisfied.

The HINT comparison protocol and physical acceptance gates remain in #302; these placement tests are not equilibrium or derivative-accuracy certificates.


rogeriojorge:

PR #423, opened 2026-09-21
Measure the supplied primal state without replacing its edge

rogeriojorge:
The draft primal certificate could identify the supplied coefficients while measuring a different state: `_primal_measurements` reassembled the fixed edge from the parameters before evaluating forces and geometry. An edge perturbation of 1e-3 was silently removed. This also made a free-boundary diagnostic appear geometrically invalid when a fixed-boundary accessor replaced its evolved edge.

Measure the supplied state directly and check fixed-edge consistency separately. Both status certificates and direct derivative admission reject mismatched edges even when `primal_tol=None`; diagnostics report the boundary error separately from geometry. Explicit free-boundary inputs are rejected because this accessor lacks the coupled field and constraint data. The boundary check uses the existing boundary transform without rebuilding profiles.

Stacked on #422 (`e25aa1fd`), following #421, #418 and #417. Three existing files change; no new files or equilibrium/response tolerance changes. The callback argument unpack is also made explicit to satisfy the type checker on this stack.

Validation and limits:

- The supplied-edge regression fails on `e25aa1fd`: the parent replaces the measured edge. It passes with this change.
- Tests cover the canonical edge, an intentionally perturbed edge, eager and staged direct admission, and unsupported free-boundary input.
- A real JIT-enabled CPU Solovev solve (NS5, MPOL3) returned status 0, zero boundary error, projected residual 5.01e-16, raw residual 1.03e-15 and FSQ 1.16e-29.
- GPU and full optimization qualification remain open. This is an exact-state/admission correction, not an observable-accuracy or speedup claim.

Initial validation at `c5fa14d8`: 55 passed, 11 hardware/device-count skips across the optimizer-penalty and device modules; 11 additional cases passed with two forced CPU devices. All 12 then-changed executable source lines were covered. These counts precede the compatibility follow-up below. The tests initialize copied configurations before tracing, matching the production setup/cache contract.

```bash
python -m coverage run --branch -m pytest -q tests/test_optimize_penalty.py tests/test_device.py
XLA_FLAGS=--xla_force_host_platform_device_count=2 OMP_NUM_THREADS=1 OPENBLAS_NUM_THREADS=1 PYTHONPATH=. python -m pytest tests/test_device.py -q -k 'host_callback_refinement_follows or status_certificate_preserves'
```

Compatibility follow-up at `02bf33d9`: compare the supplied edge with a roundoff allowance of 32 machine epsilons times the largest expected R/Z edge coefficient, and report that allowance separately. Host NumPy and staged JAX transforms can differ slightly without changing the physical boundary. This does not relax primal or linear-response tolerances and never replaces measured coefficients. Non-finite errors and material mismatches remain rejected. Align parameters, state and mask with the explicit configuration device before either public measurement.

Follow-up validation: eight focused cases pass with two forced CPU devices, including explicitly enabled JIT, one-ulp/material edge preservation, roundoff acceptance and material rejection at two geometry scales, native asymmetric setup, and real public measurement from the other device. Coverage includes all nine newly changed executable source statements; aggregate changed-test/source coverage is 98% (the unavailable-device skip is unexecuted). The old exact gate at `c5fa14d8` rejects the one-ulp perturbation. Ruff, mypy and whitespace checks pass. These checks do not qualify accelerator execution or a full optimizer.

Combined regression on the proposed minimum integration environment: all 71 tests in `tests/test_optimize_penalty.py` and `tests/test_device.py` passed with two forced CPU devices (183.87 s, no skips). The device regression and staged boundary tests explicitly enable JIT; other tests retain their existing execution policy. This is validation timing, not a performance comparison.

Combined-run environment: Python 3.11.14, JAX/jaxlib 0.10.1, NumPy 2.4.6, SciPy 1.16.0, SOLVAX 0.21.0.

```bash
XLA_FLAGS=--xla_force_host_platform_device_count=2 OMP_NUM_THREADS=1 OPENBLAS_NUM_THREADS=1 PYTHONPATH=. python -m pytest -q tests/test_optimize_penalty.py tests/test_device.py
```


rogeriojorge:
### Explicit CPU/GPU callback certification on the current stack

At `02bf33d9141429f19e92a67db28e6a9b2dd4b7df`, a JIT-enabled NS=5, MPOL=3 Solovev check passed with explicit CPU and CUDA configurations. Python 3.12.13, JAX/jaxlib 0.11.1, NumPy 2.5.3, SciPy 1.18.1, SOLVAX 0.22.0; NVIDIA RTX A4000, float64, four CPU threads, GPU preallocation disabled.

| configured device | status | projected residual | raw residual | boundary error | boundary allowance |
|---|---:|---:|---:|---:|---:|
| CPU | 0 | 9.5972e-16 | 3.5735e-15 | 2.4323e-17 | 2.8415e-14 |
| CUDA | 0 | 4.5163e-16 | 1.4307e-15 | 0 | 2.8415e-14 |

Both report valid geometry, derivative admission, and attainment of the requested refinement tolerance. Re-measuring the returned host coefficients reproduces each certificate exactly; CPU/CUDA coefficients agree to a maximum absolute difference of 3.1641e-15. Template device identities were checked against the explicit configurations. The CPU's nonzero edge error is a real solved-state example that the former exact-equality gate would reject.

The measured callback costs were 15.40 s and 27.22 s respectively, including first-use work in one process. This is a correctness check, not an isolated cold-start or speedup benchmark. The native result's array placement was not separately recorded before host conversion. `primal_tol=None` means neither certificate claims a strict observable-specific root tolerance. Full GPU optimization, asymmetric accelerator cases and larger resolutions remain open.

<details><summary>Reproducer (run from the pinned repository root)</summary>

Set `CUDA_VISIBLE_DEVICES=0`, `XLA_PYTHON_CLIENT_PREALLOCATE=false`, `OMP_NUM_THREADS=4`, `OPENBLAS_NUM_THREADS=4`, and `PYTHONPATH=.`; run the script below with a 180-second process limit in an environment containing the versions above.

```python
import dataclasses
import hashlib
import json
import time
from pathlib import Path
import jax
import numpy as np
from vmex.core import implicit as im
from vmex.core.input import VmecInput
jax.config.update('jax_enable_x64', True)
source = VmecInput.from_file('examples/data/input.solovev')
inp = dataclasses.replace(source, mpol=3, ns_array=np.array([5]),
    rbc=source.rbc[:,:3], rbs=source.rbs[:,:3], zbc=source.zbc[:,:3], zbs=source.zbs[:,:3])
results = []
states = []
def emit(row):
    print(json.dumps(row, allow_nan=False), flush=True)
for platform in ('cpu', 'gpu'):
    device = jax.devices(platform)[0]
    start = time.monotonic()
    emit({'phase':'start','platform':platform,'device':str(device), 'jax':jax.__version__})
    cfg = im.make_config(inp, device=device, ftol=1e-10, max_iterations=400)
    params = im.params_from_input(inp, device=device)
    emit({'phase':'config_ready','platform':platform,'elapsed':time.monotonic()-start})
    state_np, mask_np, status, fsq, ratio = im._host_solve_and_mask_status(cfg, jax.tree.map(np.asarray, params))
    certificate = im._LAST_PRIMAL_CERTIFICATE[cfg][2]
    placement = im._params_committed_device((im._template_runtime(cfg).setup.s_full,))
    row = {'platform':platform,'elapsed':time.monotonic()-start,'status':int(status),
        'template_device':str(placement),'certificate':certificate}
    emit(row)
    results.append(row)
    states.append(state_np)
    np.savez(f'{platform}-state.npz', **{name:np.asarray(getattr(state_np,name)) for name in ('R_cos','R_sin','Z_cos','Z_sin','L_cos','L_sin')})
    assert int(status)==0 and placement==device
    assert certificate['derivative_admitted'] and certificate['primal_boundary_consistent']
    # Measure unchanged host coefficients on the configured device again.
    repeated = im.measure_primal_state(params, state_np, mask_np, cfg)
    assert repeated == certificate
for a,b in zip(jax.tree.leaves(states[0]),jax.tree.leaves(states[1])):
    np.testing.assert_allclose(a,b,rtol=1e-8,atol=1e-10)
record = {'source':'02bf33d9141429f19e92a67db28e6a9b2dd4b7df','results':results,
          'coefficient_max_difference':max(float(np.max(np.abs(a-b))) for a,b in zip(jax.tree.leaves(states[0]),jax.tree.leaves(states[1])))}
Path('result.json').write_text(json.dumps(record, indent=2, allow_nan=False)+'\n')
emit({'phase':'complete',**record})

```
</details>

rogeriojorge:

